PR #4, opened 2020-11-05
Patch/cra

msutkowski:


phryneas:
Somehow that has killed it locally.
Moved react & co to peerDepenencies, that should do it...

phryneas:
Okay, now it is missing the target codesandbox? Confused :/

PR #5, opened 2020-11-06
Add svelte example, maybe batch

msutkowski:
Left comments about the roadblocks I ran into. Maybe we can find a svelte 🧙 to assist?! This works though... and it's actually kinda nice. I never want to write another `useEffect` again 🤯


examples/svelte-example/rollup.config.js
```diff
@@ -0,0 +1,91 @@
+import svelte from 'rollup-plugin-svelte';
+import resolve from '@rollup/plugin-node-resolve';
+import commonjs from '@rollup/plugin-commonjs';
+import livereload from 'rollup-plugin-livereload';
+import { terser } from 'rollup-plugin-terser';
+import sveltePreprocess from 'svelte-preprocess';
+import typescript from '@rollup/plugin-typescript';
+import injectProcessEnv from 'rollup-plugin-inject-process-env';
```
msutkowski, line 8: This is necessary to be compatible with redux

examples/svelte-example/rollup.config.js
```diff
@@ -0,0 +1,91 @@
...
+        sourcemap: true,
+        format: 'iife',
+        name: 'app',
+        file: 'public/build/bundle.js',
+        globals: {
+            react: () => {},
+            'react-redux': () => {},
+        },
+    },
+    external: ['react', 'react-redux'],
```
msutkowski, line 45: Had to do this to bail out of the dep inclusions... but this may not be right. Who knows how configs work? 🤦‍♂️ 💣

examples/svelte-example/src/App.svelte
```diff
@@ -0,0 +1,61 @@
...
+    let data;
+    let status;
+    let error;
+
+    const selector = counterApi.selectors.query.getCount();
+    const unsubscribeSelector = store.subscribe(() => ({ data, status, error } = selector(store.getState())));
+
+    const increment = async () => ({ payload: data } = await store.dispatch(incrementCount(1)));
+
+    const decrement = async () => ({ payload: data } = await store.dispatch(decrementCount(1)));
```
msutkowski, line 18: The invalidates wasn't causing a refetch here, so i'm just reassigning it with the response. Note: the payload didn't have the type come through even though it knows about `CountResponse`. In general, I have no idea if this is the right thing to do 🙃

examples/svelte-example/src/services/counter.ts
```diff
@@ -0,0 +1,57 @@
...
+        }),
+        incrementCount: build.mutation<CountResponse, number>({
+            query(amount) {
+                return {
+                    queryString: `increment`,
+                    method: 'PUT',
+                    body: JSON.stringify({ amount }),
+                };
+            },
+            invalidates: [{ type: 'Counter' }],
```
msutkowski, line 44: As mentioned, these weren't causing an invalidation to happen on the query.

examples/svelte-example/src/App.svelte
msutkowski: Okay - I realized I'm a big ol' dumbdumb and _literally_ no-op'd the code that runs the updates 🤦

examples/svelte-example/src/services/counter.ts
msutkowski: Fixed. Disregard.

msutkowski:
Alright, I did a small cleanup pass on the examples. More importantly, I changed the signature for `fetchBaseQuery` to just default to an empty string. I think this is a good idea, but if you don't agree, I'll change it back (or youc an just override me there :P)

phryneas:
This is pretty cool :)

PR #7, opened 2020-11-07
Enhance/fetch base query

themindoverall:
This is still a WIP, but it's getting there. Will add further notes tomorrow.


src/utils/isAbsoluteUrl.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  // Don't match Windows paths `c:\`
+  if (/^[a-zA-Z]:\\/.test(url)) {
+    return false;
+  }
+
+  // Scheme: https://tools.ietf.org/html/rfc3986#section-3.1
+  // Absolute URL: https://tools.ietf.org/html/rfc3986#section-4.3
+  return /^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(url);
```
phryneas, line 15: I'd argue that not the presence of a schema, but in http/https context (and I guess we can pretty safely stay with those for a simple fetch helper), the `//` makes an url absolute. So either `://` or leading `//` (which would not be caught by this, but would define an absolute url in the current scheme).
```suggestion
  return new RegExp(`(^|:)//`).test(url)
```

src/utils/isPlainObject.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * @param {any} obj The object to inspect.
+ * @returns {boolean} True if the argument appears to be a plain object.
+ */
+export function isPlainObject(obj: any) {
```
phryneas, line 5: It's a shame that `immer` doesn't export their `isPlainObject` method, but maybe we could copy that over? It's a lot shorter.

https://github.com/immerjs/immer/blob/9007be0a68e8c521d23657729036d2eee9f5df83/src/utils/common.ts#L39-L44

src/fetchBaseQuery.ts
```diff
@@ -1,28 +1,73 @@
...
+
+    if (body && isPlainObject(body) && isJsonContentType(config.headers)) {
+      config.body = JSON.stringify(body);
+    }
+
+    url = joinUrls(baseUrl, url);
+
+    if (params) {
+      const searchParams = new URLSearchParams(params);
+      url += `?${searchParams.toString()}`;
```
phryneas, line 50: This will lead to a double `?` if the `url` already has search params.
Maybe something like this?
```suggestion
      let parsedUrl = new URL(url);
      for ([k,v] of Object.entries(params)) { url.searchParams.set(k,v) }
      url = parsedUrl.toString()
```

examples/svelte-counter/src/services/counter.ts
```diff
@@ -10,22 +10,22 @@ export const counterApi = createApi({
...
         }),
         incrementCount: build.mutation<CountResponse, number>({
-            query: (amount) => ({
-                url: `increment`,
-                method: 'PUT',
-                body: JSON.stringify({ amount }),
+            query: (body) => ({
+                url: `/increment`,
+                method: 'PUT', 
+                body, // Very clever!
```
phryneas, line 20: But won't increment and decrement behave differently now? One POSTING `5` and the other `{amount:5}`

src/fetchBaseQuery.ts
```diff
@@ -1,28 +1,73 @@
...
 interface FetchArgs extends RequestInit {
   url: string;
+  params?: URLSearchParams | Record<string, string>;
+  body?: any;
+  responseHandler?: 'json' | 'text' | ((response: Response) => Promise<any>);
+  validateStatus?: (response: Response, body: any) => boolean;
 }
 
-export function fetchBaseQuery({ baseUrl }: { baseUrl: string } = { baseUrl: '' }) {
+const defaultValidateStatus = (response: Response) => response.status >= 200 && response.status <= 299;
```
phryneas, line 12: I just jotted this down quickly, but it would be worth to check what fetch does in a 3xx case. Follow the redirect? If so, return the redirect target's return code?

src/fetchBaseQuery.ts
```diff
@@ -1,28 +1,73 @@
...
+
+const isJsonContentType = (headers: Headers) => headers.get('content-type')?.trim()?.startsWith('application/json');
```
phryneas, line 14: Probably should also handle there I had forgotten about:

application/ld+json
application/vnd.api+json

phryneas:
looking good so far, nothing to critizise really, I just commented on some nitpicks :)

examples/svelte-counter/src/services/counter.ts
msutkowski: Yea, good catch. We committed while discussing & testing random things 🎉

src/utils/isPlainObject.ts
markerikson: Use the one we have in RTK / React-Redux

src/utils/isAbsoluteUrl.ts
msutkowski: I switched it over to this, but I'll feel more comfortable once I get some more tests added in. The ones i'm doing in the counter example aren't exactly ideal :)

src/utils/isPlainObject.ts
msutkowski: Thanks @markerikson ! I've updated to this, and opened a PR in RTK to export this so we can just use that. It's also debatable if we even need to do anything other than `typeof body === 'object'`, but I'd rather have more guardrails if the cost isn't really measurable.

src/fetchBaseQuery.ts
msutkowski: I've updated this to behave like `axios`/`redaxios` do. The problem with `new URL` is that if you give it a relative url, it requires a baseUrl, which is a major headache especially if someone tried to use this with SSR? We experimented with it yesterday and it was a ton of bloat to get something that didn't even cover every case :(

src/fetchBaseQuery.ts
msutkowski: Fetch follows redirect by default 🎉

src/fetchBaseQuery.ts
msutkowski: Just leaving a note for posterity: this would only matter if we used this to inspect the headers on the response, which we currently do not do. We did discuss having an 'auto' option for `responseHandler` that would make a best guess based on the incoming response header, but as of right now that's probably overkill.

PR #8, opened 2020-11-08
Allow a shorthand of entityType for invalidates

msutkowski:
- [ ] Investigate `provides` behavior when using the shorthand version. Passing `{ type: 'Counter' }` works fine with an `invalidates: ['Counter']`, but a short version of provides does not.
- [x] Show dynamic counters and refetching
- [x] Allow for `invalidates: ['Counter']`

### DIFF
--- a/src/buildMiddleware.ts
+++ b/src/buildMiddleware.ts
@@ -38,12 +38,10 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
         endpointDefinitions[action.meta.arg.endpoint].invalidates,
         action.payload,
         action.meta.arg.arg
-        // state.mutations[action.meta.arg.endpoint]?.[action.meta.requestId]?.arg,
       );
       const toInvalidate: { [endpoint: string]: Set<string> } = {};
       for (const entity of invalidateEntities) {
-        const entityType = typeof entity === 'string' ? entity : entity.type;
-        const provided = state.provided[entityType];
+        const provided = state.provided[entity.type];
         if (!provided) {
           continue;
         }
--- a/src/buildMiddleware.ts
+++ b/src/buildMiddleware.ts
@@ -41,7 +41,8 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
       );
       const toInvalidate: { [endpoint: string]: Set<string> } = {};
       for (const entity of invalidateEntities) {
-        const provided = state.provided[entity.type];
+        const entityType = typeof entity === 'string' ? entity : entity.type;
+        const provided = state.provided[entityType];
         if (!provided) {
           continue;
         }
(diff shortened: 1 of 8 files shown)


src/buildMiddleware.ts
```diff
@@ -37,11 +37,13 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
       const invalidateEntities = calculateProvidedBy(
         endpointDefinitions[action.meta.arg.endpoint].invalidates,
         action.payload,
-        state.mutations[action.meta.arg.endpoint]?.[action.meta.requestId]?.arg
+        action.meta.arg.arg
+        // state.mutations[action.meta.arg.endpoint]?.[action.meta.requestId]?.arg,
```
msutkowski, line 41: I think this might only work when tracked is enabled? But either way, this should just reference the arg from the fulfilled action.meta, right?

msutkowski:
@phryneas It looks like we'll still need to fix `provides: ['Counter']`. If you set `getCount` to that, get/increment/decrement don't behave properly, but the `ById` versions do. Will investigate further tomorrow unless you wanna check it out.

src/buildMiddleware.ts
phryneas: very good point.

phryneas:
What about these code changes?

---

As for provides & invalidates, it is a bit of a question, what should invalidate what

Right now, I've chosen this behaviour:

<table>
<tr>
	<td>provided:
	<td>'Counter'
	<td>{type: 'Counter', id: 1}
	<td>{type: 'Counter', id: 2}
<tr>
	<td>invalidate 'Counter'
	<td>invalidated
	<td>invalidated
	<td>invalidated
<tr>
	<td>invalidate {type: 'Counter', id: 1}
	<td>
	<td>invalidated
	<td>
<tr>
	<td>invalidate {type: 'Counter', id: 2}
	<td>
	<td>
	<td>invalidated
</table>

An alternative behaviour could be this:


<table>
<tr>
	<td>provided:
	<td>'Counter'
	<td>{type: 'Counter', id: 1}
	<td>{type: 'Counter', id: 2}
<tr>
	<td>invalidate 'Counter'
	<td>invalidated
	<td>invalidated
	<td>invalidated
<tr>
	<td>invalidate {type: 'Counter', id: 1}
	<td>invalidated
	<td>invalidated
	<td>
<tr>
	<td>invalidate {type: 'Counter', id: 2}
	<td>invalidated
	<td>
	<td>invalidated
</table>

but I'm not really sure which one would be more "expected" & useful. That's really open for discussion at this point.

msutkowski:
> What about these code changes?

I think the current behavior is most likely ideal. I like the ability to invalidate everything, and for that, we should probably conside. an operation like `{ type: 'Counter', id: '*' }`. Then again, I might imagine if someone just passes the entity type by itself, that is the intention of it anyways... not sure :)

PR #9, opened 2020-11-08
basic docs structure

phryneas:
@msutkowski let's work on docs& identify TODOs in here


docs/1_basics.md
```diff
@@ -0,0 +1,78 @@
...
+
+Step 1: Create your api:
+
+```ts
+import {createApi} from '@rtk-incubator/simple-query'
+import {ListResponse, User} from './types'
+
+export const api = createApi({
+  reducerPath: 'testApi',
+  baseQuery: fetchBaseQuery({ baseUrl: 'https://reqres.in/api' }),
```
markerikson, line 11: Needs to import this, I assume

docs/1_basics.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+Step 3: use the hooks in your React components
+
+```tsx
+import {api} from './api'
+import {User} from './types'
+
+function UsersList() {
+    const { status, data } = api.hooks.getUsers.useQuery();
```
markerikson, line 50: Random non-docs observations:

- the whole `api.hooks.x.useQuery` thing feels a bit verbose. Any way we can shorten this? (yeah, I realize they can always export the pieces they care about)
- Can we throw in auto-derived booleans for `isLoading`, etc?  David K had a couple tweets to that effect recently (ie, _deriving_ booleans is fine): https://twitter.com/DavidKPiano/status/1318612363499720707

phryneas: Problem is that hooks have to start with `use`, so without renaming it will always either be `endpoint.useQuery` or `endpoint.useMutation`. With TS 4.1, that can be renamed type-safely, but so far that is a RC and prettier has still not released a version compatible with that.
Once it does, I'm planning to have `hooks.useEndpointQuery` & `hooks.endpoint.useQuery` in there side-by-side for both <4.1 and >=4.1 users.

msutkowski:
@phryneas Closing this for now - I've added the comparison table to the `docs/base` branch and converted any outstanding todos into issues.

PR #11, opened 2020-11-09
Subscription options 2

msutkowski:
Ahhh, i messed this up because I rebased. I'll just leave some comments and open a better PR tomorrow 🤦

### DIFF
--- a/src/buildMiddleware.ts
+++ b/src/buildMiddleware.ts
@@ -1,96 +1,165 @@
-import { AnyAction, AsyncThunk, Middleware, ThunkDispatch } from '@reduxjs/toolkit';
+import { AnyAction, AsyncThunk, Middleware, MiddlewareAPI, ThunkDispatch } from '@reduxjs/toolkit';
 import { batch as reactBatch } from 'react-redux';
-import { QueryState, QueryStatus, RootState } from './apiState';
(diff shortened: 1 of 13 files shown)


src/buildMiddleware.ts
```diff
@@ -1,98 +1,167 @@
...
+      for (const invalidate of invalidateSubscriptions) {
+        (toInvalidate[invalidate.endpoint] ??= new Set()).add(invalidate.serializedQueryArgs);
+      }
+    }
+
+    batch(() => {
+      for (const [endpoint, collectedArgs] of Object.entries(toInvalidate)) {
+        for (const serializedQueryArgs of collectedArgs) {
+          const querySubState = (state.queries as QueryState<any>)[endpoint]?.[serializedQueryArgs];
+          // const querySubState = querySelectors[endpoint]?.(serializedQueryArgs)(rootState);
```
msutkowski, line 98: This breaks, because the selector calls serializeQueryParams again - causing things like "\"1"\" instead of "1".

src/buildMiddleware.ts
```diff
@@ -1,98 +1,167 @@
...
+
+          if (querySubState) {
+            if (Object.keys(querySubState.subscribers).length === 0) {
```
msutkowski, line 101: This is an object now, so we have to .keys it to match previous behavior

PR #12, opened 2020-11-09
Silly useQuery/useMutation implementation

msutkowski:
This is just a test to 'see how it feels'. I don't know if I'd like this approach... but i think it could be TS-friendly? I think the better thing would probably be something like `useQuery(api, queries => queries.method(opts))`

### DIFF
--- a/examples/posts-and-counter/src/features/counter/Counter.tsx
+++ b/examples/posts-and-counter/src/features/counter/Counter.tsx
@@ -5,10 +5,9 @@ import { useQuery, useMutation } from '@rtk-incubator/simple-query/dist';
 
 export function Counter() {
   // const { data } = counterApi.hooks.getCount.useQuery();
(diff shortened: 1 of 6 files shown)


msutkowski:
This is as far as I'll be able to get this, I think. I'm also not sure if it's worth getting any further - I like the idea of just being able to `useQuery` without exporting all of the methods first, but then again, that isn't the end of the world. In regards to this, I've got no idea of how to get the `EndpointDefinitions` types to be real when calling `buildXHook` outside of `createApi`. I'm pretty sure that's the issue... I also tried to switch to calling the service method like `service.hooks[method].useQuery()` because that would for sure have the right return type already, but I'm not exactly sure on how to make that happen without dropping the types.

msutkowski:
Closing due to being an irrelevant thing once #22 is in.

PR #14, opened 2020-11-10
Fixes, add polling to svelte example

msutkowski:
Alright, clean diff from the original one from yesterday. I'll see if I get a little time to fix this later.

If you just check the example, you'll see some fun behavior happen with the polls and subscriptions that sit in state :) _(will remove this if i get to fix it)_


msutkowski:
Closing as it's wrapped into #16

PR #16, opened 2020-11-10
Flatten state

phryneas:
building on top, again.

While it does make the code a lot(!) simpler, it removes the information through nesting. Some actions need to know about the endpoint a certain substate was created for.
So I'll probably re-add that as a property on the sub-state. Which then leaves the risk of two different endpoints overwriting each other.  So the cache keys need to be prefixed with the endpoint again :see_no_evil:


phryneas:
Let's just get all of this back into `main` before we continue on a branch-on-a-branch

PR #21, opened 2020-11-13
Docusaurus

msutkowski:
Major WIP 😅


markerikson:
where you planning on hosting this?  just grab a Netlify-labeled subdomain or something?  Can set one up tied to the Redux Netlify team if you want.

msutkowski:
@markerikson Exactly. I setup Netlify and such last night, and was going to do the `.js.org` thing later today. If you wanna help out with that, I'd appreciate it because the goal was just to get it up -> transition over to your accounts as an offical redux/rtk thing when you wanted.

markerikson:
@msutkowski I would actually vote that we _not_ bother with a `.js.org` name for now.  Takes some time to add one of those, and assuming this new API works out, eventually we'd want to migrate the content into the actual RTK docs.

I'd say for this "incubator" stuff, it's better to just go with a generic Netlify-hosted name, like `rtk-query-docs.netlify.app` or something.

msutkowski:
Sounds good to me. I'm assuming you've already got the OSS Netlify account, so if you want to add it, that'd be great.

markerikson:
okay, set up the Netlify site.  

Not sure what the build command and output folder are atm.  You may want to go copy-paste a `netlify.toml` file from one of the Redux repos, paste it in here, and edit the settings - I believe that config file overrides whatever build settings are in the Netlify dashboard.

markerikson:
aaaand we have build preview!

https://deploy-preview-21--rtk-query-docs.netlify.app/

markerikson:
We should get the TS/JS conversion plugin ported over from the RTK docs.

Also, use the new Markdown code block title syntax for code block filenames, instead of comments, like:

```markdown

```js title="src/someFile.ts"

```

markerikson:
Let's merge this in and iterate on it.

PR #23, opened 2020-11-13
Fix posts example

msutkowski:


msutkowski:
Closing - will swap the API back to the previous signature in a future branch.

PR #25, opened 2020-11-13
fix `args` argument to `calculateProvidedBy`

phryneas:
Yeah... :sweat_smile: My fault :tada: 

@msutkowski can you take a look?

### DIFF
--- a/src/apiState.ts
+++ b/src/apiState.ts
@@ -4,6 +4,7 @@ import {
   EndpointDefinitions,
   BaseEndpointDefinition,
   ResultTypeFrom,
+  QueryArgFrom,
 } from './endpointDefinitions';
 import { Id, WithRequiredProp } from './tsHelpers';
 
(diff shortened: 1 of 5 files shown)


src/buildSlice.ts
```diff
@@ -145,7 +147,7 @@ export function buildSlice({
       builder
         .addCase(queryThunk.fulfilled, (draft, { payload, meta: { arg } }) => {
           const { endpoint, queryCacheKey } = arg;
-          const providedEntities = calculateProvidedBy(definitions[endpoint].provides, payload, arg.internalQueryArgs);
+          const providedEntities = calculateProvidedBy(definitions[endpoint].provides, payload, arg.originalArgs);
```
phryneas, line 150: this was the root of the problem

src/apiState.ts
```diff
@@ -28,6 +29,7 @@ export type MutationKeys<Definitions extends EndpointDefinitions> = {
 }[keyof Definitions];
 
 type BaseQuerySubState<D extends BaseEndpointDefinition<any, any, any>> = {
+  originalArgs: QueryArgFrom<D>;
```
msutkowski, line 32: This looks good to me, and I think is probably a good name to stick with. As some possible alternatives for thought: `functionArgs`, `thunkArgs`, `userArgs`, or `givenArgs`.

src/buildSlice.ts
msutkowski: Yup, that'd do it!

msutkowski (approved):

PR #29, opened 2020-11-15
Docs updates

msutkowski:
WIP 🎉


docs/concepts/queries.md
```diff
@@ -0,0 +1,51 @@
+---
+id: queries
+title: Queries
+sidebar_label: Queries
+hide_title: true
+---
+
+# `Queries`
+
+This is the most basic feature of `rtk-query`. A `query` can be performed with any data fetching library of your choice and will accept any HTTP method, but we recommend that you only use queries for `GET` requests. For anything that alters data on the server or will possibly invalidate the cache, you should use a [Mutation](./mutations).
```
phryneas, line 10: I wouldn't keep this too focused on http here. 
Theroetically, these could also be used to make grpc calls or anything else, really.

website/docusaurus.config.js
```diff
@@ -3,8 +3,8 @@ const { resolve } = require('path');
 const { linkDocblocks, transpileCodeblocks } = require('remark-typescript-tools');
 
 module.exports = {
-  title: 'RTK Query',
-  tagline: 'An advanced data fetching & caching library built on redux-toolkit',
+  title: 'Powerful data fetching and caching capabilities for every everyone',
```
markerikson: "every everyone"?  not just _some_ everyones? :)

msutkowski: I have no idea what I'm doing 😆 . I suck at marketing copy, but I think it's getting better.

docs/concepts/polling.md
```diff
@@ -1,8 +1,51 @@
...
+};
+```
+
+In an action creator without React Hooks:
+
+```ts
+const { data, status, error, refetch } = store.dispatch(actions.getCountById(id, { subscriptionOptions: { pollingInterval: 3000}}))
+```
+:::note Usage without React Hooks
+If you use polling without the convience of React Hooks, you will need to manually call `updateSubscriptionOptions` on the promise ref to update the interval. This approach varies by framework but is possible everywhere. See the [Svelte Example](../examples/svelte) for one possibility.
```
markerikson, line 37:
```suggestion
If you use polling without the convenience of React Hooks, you will need to manually call `updateSubscriptionOptions` on the promise ref to update the interval. This approach varies by framework but is possible everywhere. See the [Svelte Example](../examples/svelte) for one possibility.
```

docs/api/fetchBaseQuery.md
```diff
@@ -0,0 +1,147 @@
...
+By default, we'll assume that every `Response` you get will be parsed as `json`. In the event that you don't want that to happen, you can specify an alternative response handler like `text`, or take complete control and use a custom function that takes the raw `Response` object &mdash; allowing you to use any [`Body` method](https://developer.mozilla.org/en-US/docs/Web/API/Body).
+
+```ts title="Parse a response as formData"
+export const customApi = createApi({
+  baseQuery: fetchBaseQuery({ baseUrl: '/api/' }),
+  endpoints: (builder) => ({
+    getUsers: builder.query({
+      query: () => ({
+        url: `users`,
+        responseHandler: (response) => response.formData(),
```
phryneas, line 125: this would lead to a non-serializable value in the resolved action, so we should probably pick a different example (if there are other things as text & json?)

msutkowski: Yea, that's definitely something to consider. I'm assuming we should add a note about serialized values and store configuration? Could use some ideas on the right approach there.

docs/api/fetchBaseQuery.md
```diff
@@ -130,8 +130,7 @@ export const customApi = createApi({
 ```
 
 ### Handling non-standard Response status codes
-
-By default, we will `reject` any `Response` that has a status code that is `< 200` or `> 299` and set it to an error. This is the same behavior you've most likely experienced with `axios` and other popular libraries. In the event that you have a non-standard API you're dealing with, you can use the `validateStatus` option to customize this behavior.
+By default, we will `reject` any `Response` that does not have a status code of `2xx` and set it to an error This is the same behavior you've most likely experienced with `axios` and other popular libraries. In the event that you have a non-standard API you're dealing with, you can use the `validateStatus` option to customize this behavior.
```
markerikson, line 134: :hand: I'd prefer to change all uses of "we" to "RTK Query" or "the library" - ie, describe it in terms of the _code_ doing things, not "us" doing things.

msutkowski: I think I covered the majority of it. I'll do another pass in the future and have made a mental note to skip out on the `we` everywhere :D

PR #32, opened 2020-11-15
make `reducerPath` and `entityTypes` optional

phryneas:


msutkowski (approved):
This looks good to me. 🚀

markerikson:
Unrelated side note: since we've got bundle size analysis going on now, it would be good to do a check of the lib contents at some point and get a sense of what's actually taking up space in the bundle and what we could maybe shrink, if anything.

I think we're in the same ballpark as other data fetching libs, and I don't expect there's huge savings to be found, but every byte helps! :)

phryneas:
Then let's get this back in :(

PR #34, opened 2020-11-16
Rename postProcess to transformResponse

### LINKED ISSUE #31: postProcess naming
For sake of familiarity, we should rename `postProcess` to `transformResponse` or `responseTransformer`. Or maybe there's a better name there that is more clear?

@markerikson @phryneas thoughts?

msutkowski:
Closes #31

### DIFF
--- a/src/buildThunks.ts
+++ b/src/buildThunks.ts
@@ -40,10 +40,9 @@ export function buildThunks<InternalQueryArgs, ReducerPath extends string>({
     { state: InternalRootState<ReducerPath> }
   >(
     `${reducerPath}/executeQuery`,
-    (arg, { signal, rejectWithValue }) => {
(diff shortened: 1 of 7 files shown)


src/buildThunks.ts
```diff
@@ -42,7 +42,7 @@ export function buildThunks<InternalQueryArgs, ReducerPath extends string>({
     `${reducerPath}/executeQuery`,
     (arg, { signal, rejectWithValue }) => {
       return baseQuery(arg.internalQueryArgs, { signal, rejectWithValue }).then(
```
phryneas, line 44: On a general note: we should not use `.then` here, but change the whole method to `async` and `await` the result.
In the case the `baseQuery` would return synchronously, this would otherwise crash with "no property `.then` on ...".
While you're already here, can you do that?

phryneas (approved):

PR #38, opened 2020-11-16
query arg safeguards

phryneas:
* sort object keys before stringifying in `defaultSerializeQueryArgs`
* use "shallow stable" `arg` value in `useQuery` hooks to prevent rerenders when calling
```js
useEndpointQuery({ name: "Tom" });
```
instead of
```js
const arg = useMemo(() => ({ name: "Tom" }));
useEndpointQuery(arg);
```

a deep-equals would probably be too much, but a shallow equal by default seems sensible here


src/buildHooks.ts
```diff
@@ -58,32 +59,34 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
   return { buildQueryHook, buildMutationHook };
 
   function buildQueryHook(name: string): QueryHook<any> {
-    const startQuery = queryActions[name];
-    const querySelector = querySelectors[name];
```
phryneas, line 62: I had to move these inside to be able to spy on them in tests :facepalm:

src/utils/shallowEqual.ts
```diff
@@ -0,0 +1,30 @@
...
+
+function is(x: any, y: any) {
+  if (x === y) {
+    return x !== 0 || y !== 0 || 1 / x === 1 / y;
+  } else {
+    return x !== x && y !== y;
+  }
+}
+
+export function shallowEqual(objA: any, objB: any) {
```
phryneas, line 11: This is the implementation from react-redux. Probably overkill though.
Maybe we should use an external lib for that so there's a *chance* of not duplicating code when another lib in the bundle uses the same helper?

phryneas: @markerikson do you have an opinion there?

markerikson: Kind of the age-old question: rely on a dependency and add to the potential deps size, or copy it into your codebase and maybe end up with duplicate code?

I don't have strong feelings on the "overkill" part.

I'd be inclined to have a copy-pasted version because we need it here, Redux doesn't have one, we can't import it from React-Redux even if R-R did export it, etc.

msutkowski (commented):
Looks good. I'm going to add in a quick test for the arg swap ({a: 1, b: 2} -> {b: 2, a: 1}) for you.

`Bob` and `Alice` sound like my type of people! 🍨 🧇

msutkowski (approved):

phryneas:
Then let's get this in.

PR #39, opened 2020-11-17
Add isLoading, isError, isSuccess, isFetching booleans

msutkowski:
Notes:

- `isFetching` will be `true` during any time a  request is in-flight.
- `isLoading` will only be `true` when a request is in flight until a result is returned. 

**Possible states:**
isLoading = false, isFetching = false
isLoading = true, isFetching = true
isLoading = false, isFetching = true

**Additional notes on behavior:**
As long as we have `data` for a query and it is not garbage collected (see `keepUnusedDataFor`), the cached result will be served until a future request updates it. Additionally, if a subsequent request for the same query errors, the cached result will continue to be shown until it is resolved -- even if retries are exhausted.

- [x] Add tests


msutkowski:
Being that this is so basic, I'll add tests for this after we merge in #38.

src/buildSlice.ts
```diff
@@ -40,6 +42,13 @@ function updateMutationSubstateIfExists(
   }
 }
 
+function setRequestStatusFlags(status: keyof typeof QueryStatus, substate: QuerySubState<any> | MutationSubState<any>) {
+  const flags = getRequestStatusFlags(status);
+  Object.keys(flags).forEach((key) => {
+    (substate as any)[key] = (flags as any)[key];
```
markerikson, line 48: Couldn't we just derive these on read or something, like in the selectors?

msutkowski:
How does typescript even work? There is an issue with `IS_DEV` currently and I'm not too sure of whats going on. Sometimes it throws, sometimes it doesn't.

src/index.ts
```diff
@@ -29,7 +29,10 @@ function defaultSerializeQueryArgs(args: any, endpoint: string) {
   return `${endpoint}/${JSON.stringify(args)}`;
 }
 
-const IS_DEV = () => process?.env?.NODE_ENV === 'development';
+const IS_DEV = () => {
+  return true;
+  return process?.env?.NODE_ENV === 'development';
```
msutkowski, line 34: For some reason this throws `process is undefined` randomly when running the React sample app.

src/buildSelectors.ts
```diff
@@ -62,23 +63,33 @@ export function buildSelectors<InternalQueryArgs, Definitions extends EndpointDe
...
+        }
+      : substate;
+  }
+
   function buildQuerySelector(
     endpoint: string,
     definition: QueryDefinition<any, any, any, any>
   ): QueryResultSelector<any, RootState> {
     return (arg?) => (rootState: RootState) =>
       (arg === skipSelector
```
phryneas, line 80: This could now just be wrapped as a whole, like 
```ts
      withRequestFlags(
        (arg === skipSelector
          ? undefined
          : (rootState[reducerPath] as InternalState).queries[serializeQueryArgs(definition.query(arg), endpoint)]) ??
          defaultQuerySubState
      );
```

BUT: I'd say we start to need reselect here as the result would become unstable otherwise. Mabe something like

```ts
  function selectInternalState(rootState: RootState) {
    return rootState[reducerPath] as InternalState;
  }

  function buildQuerySelector(
    endpoint: string,
    definition: QueryDefinition<any, any, any, any>
  ): QueryResultSelector<any, RootState> {
    return (arg?) => {
      const selectQuerySubState = createSelector(
        selectInternalState,
        (internalState) =>
          (arg === skipSelector
            ? undefined
            : internalState.queries[serializeQueryArgs(definition.query(arg), endpoint)]) ?? defaultQuerySubState
      );
      return createSelector(selectQuerySubState, withRequestFlags);
    };
  }
```

The problem is: we'll also need some to start using the selectors differently in `useQuery/useMutation`.

So in `buildQueryHook`, it was
```ts
      const querySelector = querySelectors[name];
      const currentState = useSelector(querySelector(skip ? skipSelector : stableArg));
```
and should probably become
```ts
      const buildQuerySelector = querySelectors[name];
      const querySelector = useMemo(() => buildQuerySelector(skip ? skipSelector : stableArg), [skip, stableArg]);
      const currentState = useSelector(querySelector);
```

same goes for the mutation stuff.

An alternative to that would be using `shallowEquals` in `useSelector` and keeping this as-is.
But long-term I *assume* that this is the better direction to go.
Maybe we should rename `selectors` to `getSelector` on the api return object?

phryneas (changes requested):

src/index.ts
phryneas: Can you try
```js
const IS_DEV = () => typeof process !== "undefined" && process.env.NODE_ENV === 'development';
```
?

examples/react/src/app/services/posts.ts
```diff
@@ -14,7 +14,10 @@ export const postApi = createApi({
   endpoints: (build) => ({
     getPosts: build.query<PostsResponse, void>({
       query: () => 'posts',
-      provides: (result) => result.map(({ id }) => ({ type: 'Posts', id })),
+      provides: (result) => [
+        ...result.map(({ id }) => ({ type: 'Posts', id } as const)),
+        { type: 'Posts', id: 'LIST' },
```
markerikson, line 19: :question: why is this "LIST" bit needed?

src/buildSelectors.ts
msutkowski: Hmm, regarding the naming, I don't really have an opinion. I think keeping it as `selectors` is fine.

examples/react/src/app/services/posts.ts
msutkowski: If you invalidate `['Posts']`, it will cause _every query_ that provides a type of 'Post' to be rerun, even if it provides a specific `id`. So in the example of this demo, the way it was set up originally was that when you added a new post, it would cause `PostDetail` to switch into a loading state, which is not what we want. So switching to 'LIST' here causes the `addPost` mutation to _only_ rerun `getPosts`, and `getPost(2)` remains stable as it's data doesn't change.

Does that make sense? :D

markerikson: so is "LIST" a magic thing for queries in general, or something specific to this particular demo?  Is this behavior documented somewhere?

phryneas: It's just an id, like 5 or 7 would be, with a descriptive name. Nothing forces you to have those `{ type: 'Posts', id: something }` correlate to anything.

We have the special situation here that adding a new element should not refetch any existing id, but that it should refetch all lists (because it might be somewhere on there). So all lists `provide` the ids they really have, plus LIST - and the add endpoint invalidates that.

Of course, you could also introduce a new type PostsList for that or something.

msutkowski: 'LIST' is just a generic id that applies to any query --  you could also call it '*', 'ALL' or 'BANANA' if you wanted. In theory, you could create multiple ids like 'SIDEBAR_LIST', 'PAGE_LIST', and if for some reason you wanted to... you could only refresh one of those two when adding a post.

It's hinted at in the docs in the [example here](https://rtk-query-docs.netlify.app/concepts/mutations#advanced-mutations-with-revalidation), but I need to add more copy spelling out the exact behavior and demonstrate further use cases.

phryneas: And yeah, these little tricks should probably make it into the advanced invalidation section. Right now, it's only mentioned in the example in the readme I believe.

src/buildSelectors.ts
phryneas: I mean, those are really more selector factories than selectors though, so I'm not really happy with just `selectors`.
@markerikson do you have an opinion there?

examples/react/src/app/services/posts.ts
markerikson: Right, that's kinda my point - I wasn't sure if it was something unique to this demo, or built into the library itself, or a fallout of something that's inherent in the lib.

msutkowski: Example scenarios

**Invalidating everything**
**queries:**
getPosts: provides ['Posts']
getPost: provides [{ type: 'Posts', id }]
**mutations:**
addPost: invalidates ['Posts']

**Component:**
<App>
<AddPost />
<PostsList />
<PostDetail id={id} />
<PostDetail id={id} />
<PostDetail id={id} />
</App>

Behavior: 
When you `addPost`, it will cause each `PostDetail` component to go into a loading state because `addPost` invalidates the root entity, which causes any query that provides 'Posts' to be re-run.


**Selectively invalidating lists**
**queries:**
getPosts: provides [{ type: 'Posts', id: 'LIST' ]
getPost: provides [{ type: 'Posts', id }]
**mutations:**
addPost: invalidates [{'Posts', id: 'LIST'}]

**Component:**
<App>
<AddPost />
<PostsList />
<PostDetail id={id} />
<PostDetail id={id} />
<PostDetail id={id} />
</App>

Behavior: 
When you `addPost`, it will only cause the `PostsList` to into a loading state because `addPost` invalidates the 'LIST' id, which causes `getPosts` to rerun (because it provides that specific id).

src/buildSelectors.ts
markerikson: Only partly following along because I'm working, so not really :)  other than "shorter is good", but also "accurate is good"

examples/react/src/app/services/posts.ts
msutkowski: And yea, docs are still a WIP, but I'll get them added in. There is also the case that this may end up not being a concern for folks if we do background data fetching and are able to differentiate between `loading` and `background fetching`. Either way, I do need to spell out the network request behavior :)

msutkowski: Okay! Now we have 'background fetching' 🎉

PR #41, opened 2020-11-17
change `InternalQueryArgs` generic to `BaseQuery` generic, add timestamps to state

phryneas:


test/createApi.test.ts
```diff
@@ -155,3 +156,131 @@ describe('wrong entityTypes log errors', () => {
...
+    api.injectEndpoints({
+      endpoints: (build) => ({
+        query: build.query<'RetVal', 'Arg'>({
+          query: (x: 'Arg') => 'From' as const,
+          transformResponse(r: 'To') {
+            return 'RetVal' as const;
+          },
+        }),
+        query1: build.query<'RetVal', 'Arg'>({
+          // @ts-expect-error
```
phryneas, line 176: this is nice, since TS 3.9 we don't need a type testing library any more :)

PR #42, opened 2020-11-18
Docs: Add more clarity to mutations, queries and createApi

msutkowski:


docs/api/createApi.md
```diff
@@ -15,22 +15,90 @@ The main point where you will define a service to use in your application.
...
+connect(null, mapStateToProps);
+```
+
+```js title="Svelte"
+$: ({ data, status } = selectors.getPosts()($store));
+```
+
+#### hooks
+
+Hooks are specifically for React Hooks users. If you are not using React Hooks, you shouldn't `export` them – this way they're not included in the bundle or pull in the React dependency.
```
phryneas, line 71: I would remove the reference to "bundle" there. Not exporting them will not reduce bundle size.

PR #43, opened 2020-11-21
Feature/use prefetch

msutkowski:
Implements `usePrefetch` and tests for every scenario I can think of at this time.


docs/concepts/prefetching.md
```diff
@@ -6,3 +6,92 @@ hide_title: true
...
 
 # `Prefetching`
+
+The goal of prefetching is to make data fetch _before_ the user navigates to a page or attempts to load some known content.
+
+There are a handful of situations that you may want to do this, but some very common use cases are:
+
+1. User hovers over a navigation element
+2. User hovers over a list element that is a link
+3. User hovers over a next pagination button
```
phryneas, line 16:
```suggestion
3. User hovers over a next pagination button
4. User navigates to a page and you know that some components down the tree will require said data. This way, you can prevent fetching waterfalls.
```

docs/concepts/prefetching.md
```diff
@@ -6,3 +6,92 @@ hide_title: true
...
+   - If specified, it will only run the query if the difference between `new Date()` and the last `fulfilledTimeStamp` is greater than the given value
+
+2. `force`
+
+   - If `force: true`, it will ignore the `ifOlderThan` value if it is set and the query will be run even if it exists in the cache.
+
+#### Behavior of the hook
+
+1. If `force: true` is set during the declaration or at the call site, the query will be run no matter what. The one exception to that is if the same query is already in-flight.
+2. If no options are specified and the query exists in the cache, the query will be performed.
```
phryneas, line 33:
```suggestion
2. If no options are specified and the query exists in the cache, the query will not be performed.
```
?

docs/concepts/prefetching.md
```diff
@@ -6,3 +6,92 @@ hide_title: true
...
+
+The goal of prefetching is to make data fetch _before_ the user navigates to a page or attempts to load some known content.
+
+There are a handful of situations that you may want to do this, but some very common use cases are:
+
+1. User hovers over a navigation element
+2. User hovers over a list element that is a link
+3. User hovers over a next pagination button
+
+### Prefetching with React Hooks
```
phryneas, line 18: This whole section (except for the example) misses to mention the fact that there is a hook *and* a callback.
It reads to me as if the hook call were already fetching.

docs/concepts/prefetching.md
```diff
@@ -6,3 +6,92 @@ hide_title: true
...
+2. `force`
+
+   - If `force: true`, it will ignore the `ifOlderThan` value if it is set and the query will be run even if it exists in the cache.
+
+#### Behavior of the hook
+
+1. If `force: true` is set during the declaration or at the call site, the query will be run no matter what. The one exception to that is if the same query is already in-flight.
+2. If no options are specified and the query exists in the cache, the query will be performed.
+3. If no options are specified and the query _does not exist_ in the cache, the query will be performed.
+   - `isLoading` will be true, `isFetching` will be true
```
phryneas, line 35: It's not really clear that this means the value returned from the `useQuery` hook used somewhere else - this could also be interpreted as if the hook would return these flags

src/buildHooks.ts
```diff
@@ -56,6 +63,15 @@ export type Hooks<Definitions extends EndpointDefinitions> = {
...
   TS41Hooks<Definitions>;
 
+export type PrefetchOptions =
+  | { force: boolean }
+  | {
+      ifOlderThan: false | number;
+    };
+
+const hasTheForce = (options: any): options is { force: boolean } => options?.force;
+const hasMaxAge = (options: any): options is { ifOlderThan: false | number } => options?.ifOlderThan;
```
phryneas, line 73: This will not test for `ifOlderThan` false.

src/buildHooks.ts
```diff
@@ -67,7 +83,51 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+  ) {
+    const dispatch = useDispatch<ThunkDispatch<any, any, AnyAction>>();
+    const store = useStore();
+    const latestStateRef = useRef(store.getState());
+
+    useLayoutEffect(() => {
+      return store.subscribe(() => {
+        latestStateRef.current = store.getState();
+      });
+    }, [store]);
```
phryneas, line 100:
```suggestion
```
Why do we need the subscription there?

src/buildHooks.ts
```diff
@@ -67,7 +83,51 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+
+    return useCallback(
+      (arg: any, opts: PrefetchOptions = { force: false, ifOlderThan: false }) => {
+        const force = hasTheForce(opts) ? opts.force : hasTheForce(options) ? options.force : false;
+        const maxAge = hasMaxAge(opts) ? opts.ifOlderThan : hasMaxAge(options) ? options.ifOlderThan : false;
+
+        const queryAction = (force: boolean = true) => queryActions[endpointName](arg, { forceRefetch: force });
+
+        const latestStateValue = querySelectors[endpointName](arg)(latestStateRef.current);
```
phryneas, line 109:
```suggestion
        const latestStateValue = querySelectors[endpointName](arg)(store.getState());
```

src/buildHooks.ts
```diff
@@ -67,7 +83,51 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          const shouldRetrigger = (Number(new Date()) - Number(new Date(lastFulfilledTs))) / 1000 >= maxAge;
+          if (shouldRetrigger) {
+            dispatch(queryAction());
+          }
+        } else {
+          // If prefetching with no options, just let it try
+          dispatch(queryAction(false));
+        }
+      },
+      [endpointName, dispatch, options]
```
phryneas, line 128:
```suggestion
      [endpointName, dispatch, options, store]
```

src/buildHooks.ts
```diff
@@ -67,7 +83,51 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+    const store = useStore();
+    const latestStateRef = useRef(store.getState());
+
+    useLayoutEffect(() => {
+      return store.subscribe(() => {
+        latestStateRef.current = store.getState();
+      });
+    }, [store]);
+
+    return useCallback(
```
phryneas, line 102: This whole callback could also be implemented as a thunk and thus also be available for non-hook-usage with all the fancy options

phryneas (commented):
I've left some review but didn't look at the tests yet.

src/buildHooks.ts
msutkowski: Skipping due to switching to prefetchThunk usage

PR #51, opened 2020-11-25
Fix fetchBaseQuery: Throw when invalid & Improve error handling

msutkowski:


msutkowski:
I'm just going to leave this here. Error handling doesn't work anything like I'd expect it to. If I get a chance later, I'll write up what I think should happen -- if not, tomorrow. As of right now, the thrown error doesn't make it into the rejected payload, and due to how `createAsyncThunk` works, you can't `catch` it. This might be an idea, but not sure if you're going to merge it: https://github.com/reduxjs/redux-toolkit/pull/812

src/fetchBaseQuery.ts
```diff
@@ -15,7 +15,7 @@ const defaultValidateStatus = (response: Response) => response.status >= 200 &&
 const isJsonContentType = (headers: Headers) => headers.get('content-type')?.trim()?.startsWith('application/json');
 
 export function fetchBaseQuery({ baseUrl }: { baseUrl?: string } = {}) {
-  return async (arg: string | FetchArgs, { signal, rejectWithValue }: QueryApi) => {
+  return async (arg: string | FetchArgs, { signal }: QueryApi) => {
```
msutkowski, line 18: After trying to use this, I can't see any reason to expose `rejectWithValue` at all - we should just handle whatever error is thrown here in `buildThunks`

msutkowski (commented):
I think this might be better, but I'd be curious as to if `refetch` should be able to be handled as a promise in a component instead of being `void`? Maybe this is offbase either way though - curious to hear thoughts.

phryneas:
> but I'd be curious as to if `refetch` should be able to be handled as a promise in a component instead of being `void`? Maybe this is offbase either way though - curious to hear thoughts.

Thing is: if a query for that serializedArgs is already running, that refetch would immediately return with a rejection. And syncing that up so it waits for that already-running query would require quite a rework of the middleware concept. So I'm putting that off for now.

PR #53, opened 2020-11-25
typed error handling with return value from `baseQuery`

phryneas:
tests TBD


test/errorHandling.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    await hookWaitFor(() => expect(result.current.isFetching).toBeFalsy());
+    expect(result.current).toEqual(
+      expect.objectContaining({
+        isLoading: false,
+        isError: false,
+        isSuccess: true,
+        data: { value: 'success' },
+        // do we want to keep this behaviour?
+        // atm., error is not cleared out
```
phryneas, line 132: we should probably clear this out when going from isError -> isSuccess, right?

msutkowski: in general, yes :)

phryneas: added this in :)

PR #57, opened 2020-11-27
Tests: fetchBaseQuery behavior

msutkowski:
- [x] params
- [x] body
- [x] prepareHeaders
  - [x] getState


msutkowski:
@phryneas I still need to add params and body usage, but if you'd like to check the `prepareHeaders` usage, that'd be great. I figure this test will turn into an example.

PR #58, opened 2020-11-28
overhaul of union types returned by selectors & hooks

phryneas:


phryneas:
ah well, let's get these in already. might be doing the same thing for mutations later though

PR #61, opened 2020-11-29
Rework build/setup process to work better across platforms

markerikson:
- Updated `build` and `prepare` commands to not call shell scripts
- Replaced `typesversions.sh` with a JS file using `shelljs`
- Moved `prepareExamples` call to a separate run script


typesversions.js
```diff
@@ -0,0 +1,21 @@
...
+  cp(file, 'dist/min-4.1/');
+}
+
+const additionalTypes = `
+import { EndpointDefinitions } from './endpointDefinitions';
+export declare type TS41Hooks<Definitions extends EndpointDefinitions> = unknown;
+export {};
+`;
+
+fs.appendFileSync('dist/min-4.1/ts41Types.d.ts', additionalTypes);
```
phryneas, line 21: Would need to completely nuke & override the existing files. Also, the target here is `dist/ts41Types.d.ts`, not `dist/min-4.1/ts41Types.d.ts`

phryneas (approved):

PR #64, opened 2020-11-29
Implement retrying

phryneas:


phryneas:
@msutkowski wanna continue on this one?

test/retryStaggered.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      endpoints: (build) => ({
+        q1: build.query({
+          query: () => {},
+        }),
+        q2: build.query({
+          query: () => {},
+          extraOptions: { maxRetries: 8 },
+        }),
+      }),
+    });
```
phryneas, line 65: @markerikson @msutkowski this is how "retry on error" would look like - that way, additional "enhancers" like that can easily be injected, for example a "fetch new token after 403 enhancer". Comments on the api?

PR #65, opened 2020-11-29
Docs content and layout editing pass

markerikson:
This PR:

- Redoes the splash screen content and layout
- Renames the "Quick Start" page to "Getting Started"
- Copies over the custom CSS and code block themes from the main Redux site
- Does a cleanup pass on the docs content


docs/concepts/mutations.md
```diff
@@ -99,21 +104,19 @@ export const PostDetail = () => {
...
+For RTK Query, _entities_ are just a name that you can give to a specific collection of data to control caching and invalidation behavior, and are defined in an `entityTypes` argument. For example, in an application that has both `Posts` and `Users`, you would define `entityTypes: ['Posts', 'Users']` when calling `createApi`.
 
-2. **Provides**
+#### Provides
 
-- A `query` can _provide_ entities to the cache.
-  - Accepts either an array of `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
+A _query_ can _provide_ entities to the cache. The `provides` argument accepts either an array of `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
```
phryneas, line 115:
```suggestion
A _query_ can _provide_ entities to the cache. The `provides` argument argument can both be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
```

docs/concepts/pagination.md
```diff
@@ -5,11 +5,11 @@ sidebar_label: Pagination
 hide_title: true
 ---
 
-# `Pagination`
+# Pagination
 
-RTK Query makes it simple to integrate with a standard index-based pagination API. This is the most common form of pagination that you'll need to implement. This is how simple it can be:
+RTK Query makes it straighforward to integrate with a standard index-based pagination API. This is the most common form of pagination that you'll need to implement.
```
msutkowski, line 10:
```suggestion
RTK Query makes it straightforward to integrate with a standard index-based pagination API. This is the most common form of pagination that you'll need to implement.
```

msutkowski (commented):
Great job @markerikson! thanks for your help on this 💯

PR #66, opened 2020-11-30
Add a basic exponential backoff, additional tests and helpers, example

msutkowski:
Sorry, I didn't get a ton of time to work on this tonight. I can wrap this up in the morning unless you beat me to it.


msutkowski (commented):
Overall, I like the composability of `retryStaggered`. Someone could do `retryStaggered(baseQuery, { maxRetries: 0})`, then set the `maxRetries` on a given endpoint... nice overall behavior & API surface.

I think we should consider making a `retryDelay` or `retryTimeout` an option. Alternatively, we could do fancy things and say 'make X requests in Y period of time', and we'll automatically _do the math_™️

PR #67, opened 2020-11-30
add apiProvider

phryneas:


src/ApiProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Api } from './apiTypes';
+
+/**
+ * Can be used as a Provider if you **do not have a Redux store**.
+ * Using this together with an existing redux store, both will
+ * conflict with each other - please use the traditional redux setup
+ * in that case.
+ */
+export function ApiProvider<A extends Api<any, {}, any, string>>(props: { children: any; api: A }) {
+  const [store] = React.useState(() =>
```
markerikson, line 13: :hand: `useMemo()` instead here?

markerikson (changes requested):

src/ApiProvider.tsx
markerikson: nm, I'm blind :) lazy init works fine here

markerikson (approved):

PR #70, opened 2020-12-01
Fix return types for matchers?

msutkowski:
I was testing out `extraReducers` with the matchers, and noticed that the return types were inaccurate. This is probably the wrong way to type this, but figured I'd kick it over.

I have a feeling this isn't the right solution though, either way.

### DIFF
--- a/examples/react/src/app/services/posts.ts
+++ b/examples/react/src/app/services/posts.ts
@@ -1,5 +1,4 @@
 import { createApi, fetchBaseQuery, retry } from '@rtk-incubator/rtk-query';
-import { setCredentials } from 'src/features/auth/authSlice';
 import { RootState } from '../store';
 
 export interface Post {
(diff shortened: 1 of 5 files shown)


examples/react/src/app/services/posts.ts
```diff
@@ -30,19 +29,20 @@ const baseQuery = fetchBaseQuery({
...
     login: build.mutation<{ token: string; user: User }, any>({
       query: (credentials: any) => ({ url: 'login', method: 'POST', body: credentials }),
-      onSuccess: (arg, mutationApi, result) => {
-        const { token, user } = result;
-        // Set our user and token in the store for future requests
-        mutationApi.dispatch(setCredentials({ token, user }));
+      extraOptions: {
+        backoff: () => {
+          // We intentionally error once on login, and this breaks out of retrying. The next login attempt will succeed.
+          retry.fail({ fake: 'error' });
```
msutkowski, line 44: This seemed like a good example for `retry.fail` usage.

msutkowski (commented):
If you fix this, can you also add in an example of passing through errors to the rejected matcher? Or typed error usage in general?

examples/react/src/app/services/posts.ts
phryneas: Wouldn't that mean that we never retry, so the whole thing is pretty pointless?

msutkowski: Kind of. Maybe this wasn't the best example 🤦 .

phryneas:
Pushed some changes. From my side this looks ready, your call if we do any more changes to the example now, or some time later

PR #90, opened 2020-12-05
Implement refetchOnFocus, refetchOnReconnect

msutkowski:
This is an idea that implements basic `refetchOnFocus` and `refetchOnReconnect` behaviors. I'm not sure if this is the best way to implement this, but wanted to throw it up here for discussion.

The goal here was to keep the config in the `querySubState` so we can leverage the middleware and support these features outside of hooks usage.

The other thing I was thinking was using the `configSlice` to have 'smarter' polling control... where we could just reject in the condition if it's offline, but keep the polling interval in-tact.


msutkowski:
Some things to think about.... we need to chart out all of the behavior between `refetchOnFocus`/`refetchOnMountOrArgChange`/`refetchOnReconnect` as well as polling. 

I was thinking about adding a configuration option to also skip requests in the condition when the browser is offline or unfocused. We can do that easily now that it's in the config state. The only potential downside there would be that we'd keep generating requests that get rejected, but I don't think it actually matters. This would be the cleanest way tho imo, as we wouldn't have to introduce any type of runner in the middleware, and it could just keep creating attempts that get rejected until conditions are valid. 

Any thoughts there @phryneas ?

docs/api/createApi.md
```diff
@@ -216,12 +219,36 @@ Defaults to `false`. This setting allows you to control whether RTK Query will o
...
 - `number` - **Value is in seconds**. If a number is provided and there is an existing query in the cache, it will compare the current time vs the last fulfilled timestamp, and only refetch if enough time has elapsed.
 
-If you specify this option as well as with `skip: true`, this **will not be evaluated** until `skip` is false.
+If you specify this option alongside `skip: true`, this **will not be evaluated** until `skip` is false.
 
 :::note
 You can set this globally in `createApi`, but you can also override the default value and have more granular control by passing `refetchOnMountOrArgChange` to each individual hook call or when dispatching the [`initiate`](#initiate) action.
 :::
 
+### `refetchOnFocus`
```
phryneas, line 228: Are we missing docs that `setupListeners` is required to use this feature? (And on `setupListeners` in general)

src/apiState.ts
```diff
@@ -81,6 +91,8 @@ type BaseQuerySubState<D extends BaseEndpointDefinition<any, any, any>> = {
   endpoint: string;
   startedTimeStamp: number;
   fulfilledTimeStamp?: number;
+  refetchOnReconnect?: boolean;
+  refetchOnFocus?: boolean;
```
phryneas, line 95:
```suggestion
```

I believe these can be removed now?

src/apiState.ts
```diff
@@ -103,6 +115,8 @@ export type QuerySubState<D extends BaseEndpointDefinition<any, any, any>> = Id<
       endpoint?: string;
       startedTimeStamp?: undefined;
       fulfilledTimeStamp?: undefined;
+      refetchOnReconnect?: undefined;
+      refetchOnFocus?: undefined;
```
phryneas, line 119:
```suggestion
```

I believe these can be removed now?

src/buildHooks.ts
```diff
@@ -118,7 +121,16 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
   }
 
   function buildQueryHook(name: string): QueryHook<any> {
-    return (arg: any, { refetchOnMountOrArgChange = false, skip = false, pollingInterval = 0 } = {}) => {
+    return (
+      arg: any,
+      {
+        refetchOnMountOrArgChange = config.refetchOnMountOrArgChange,
+        refetchOnFocus = config.refetchOnFocus,
+        refetchOnReconnect = config.refetchOnReconnect,
```
phryneas, line 129: Assuming the global config on `refetchOnMountOrArgChange` & co can be changed at some point in time using the global config state, shouldn't these default options be selected from the config state instead of the initial config object?
So, with a `useSelector` in the hook body.

Or: does it even make sense to set to the default values here instead of just applying whatever is in the config state when "reading" the value later in the middleware?

src/buildMiddleware.ts
```diff
@@ -70,11 +71,53 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
...
+
+    batch(() => {
+      for (const queryCacheKey of Object.keys(subscriptions)) {
+        const querySubState = queries[queryCacheKey];
+
+        const allSubsAreFalse = Object.values(subscriptions[queryCacheKey] || {}).every(
+          (entry) => type in entry && !entry[type]
+        );
+
+        const shouldRefetch = (baseTypeValue && !allSubsAreFalse) || !allSubsAreFalse;
```
phryneas, line 100: Hmm. `(a && b) || b` -> is that not always `b`?
Meaning the global default would be ignored at this point?

src/buildSlice.ts
```diff
@@ -236,11 +241,39 @@ export function buildSlice({
...
   });
 
+  const configSlice = createSlice({
+    name: `${reducerPath}/config`,
+    initialState: {
+      online: true,
+      focused: true,
+      refetchOnMountOrArgChange: config.refetchOnMountOrArgChange,
+      refetchOnReconnect: config.refetchOnReconnect,
+      refetchOnFocus: config.refetchOnFocus,
```
phryneas, line 251:
```suggestion
      ...config
```
is probably more future-proof and saves a few bytes

src/buildSlice.ts
```diff
@@ -236,11 +241,39 @@ export function buildSlice({
     },
   });
 
+  const configSlice = createSlice({
+    name: `${reducerPath}/config`,
+    initialState: {
+      online: true,
```
phryneas, line 247: can we assume this?

src/index.ts
```diff
@@ -38,6 +39,8 @@ export function createApi<
   endpoints,
   keepUnusedDataFor = 60,
```
phryneas, line 40: we could also add this to the config state

src/utils/isOnline.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Assumes a browser is online if `undefined`, otherwise makes a best effort
+ * @link https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine
+ */
+export function isOnline() {
+  return navigator.onLine === undefined || navigator.onLine;
```
markerikson, line 6: Is `navigator` available in a Node environment?

msutkowski: Great catch 🎉

src/apiState.ts
msutkowski: Removed

src/apiState.ts
msutkowski: Removed

src/buildHooks.ts
msutkowski: I've kept them here for now, but I'll see about removing them in a cleanup pass.

src/buildMiddleware.ts
msutkowski: Updated, but probably not in a great way. Would appreciate any cleanup help here!

src/buildSlice.ts
msutkowski: I'm using the `isOnline` util now. Behaviors are tested in `utils.test`

src/index.ts
msutkowski: Added

src/utils/isOnline.ts
msutkowski: Updated and added tests to check for the expected behaviors.

src/buildActionMaps.ts
```diff
@@ -81,7 +81,7 @@ export function buildActionMaps<Definitions extends EndpointDefinitions, Interna
   function buildQueryAction(endpoint: string, definition: QueryDefinition<any, any, any, any>) {
     const queryAction: StartQueryActionCreator<any> = (
       arg,
-      { subscribe = true, forceRefetch = false, refetchOnMountOrArgChange = false, subscriptionOptions } = {}
+      { subscribe = true, forceRefetch = false, refetchOnMountOrArgChange, subscriptionOptions } = {}
```
phryneas, line 84: Hm. Did I accidentally change the meaning of this in the non-hooks-case or is that okay? :thinking:

src/buildMiddleware.ts
```diff
@@ -70,11 +69,58 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
...
+        const querySubState = queries[queryCacheKey];
+        const subscriptionSubState = subscriptions[queryCacheKey];
+
+        if (!subscriptionSubState || !querySubState || querySubState.status === QueryStatus.uninitialized) return;
+
+        const shouldRefetch =
+          Object.values(subscriptionSubState).some((sub) => sub[type] === true) ||
+          (Object.values(subscriptionSubState).every((sub) => sub[type] === undefined) && state.config[type]);
+
+        if (shouldRefetch) {
```
phryneas, line 117: This would currently also refetch unmounted queries. Should we do that or remove those instead?

phryneas: No, that was wrong. But still a question here:

This only loops over subscribed queries, which means that it will only refetch actually mounted stuff and stuff that is not mounted is neither removed nor refreshed, potentially leading to semi-stale stuff.
So, we could either refresh those or remove them, like we do in `invalidateEntities`. Thoughts? :thinking:

msutkowski: I think we should probably refresh them.... but I could also just see the better fit for dropping them? Tough call.

phryneas: In the spirit of https://react-query.tanstack.com/docs/guides/scroll-restoration we should maybe refetch them by default, but add a config option to remove instead?
So a `refetchUnmounted` option that kicks in both here and on invalidateEntities?

Issue #89, opened 2020-12-05
Default handling for empty response body?


airjp73:
Hi all! I was trying out rtk-query by converting some old code and noticed something I thought might make sense to have some default support for.

In our app at work, all of our DELETE endpoints return a 204 with no response body. I noticed that, by default, empty responses from a request cause the mutation promise to reject. So this code always resulted in an error notification:

```js
const [deleteUser] = useDeleteUsersQuery();

const onDelete = () => {
  deleteUser(id)
    .then(unwrapResult)
    .then(/* success notification */)
    .catch(/* error notification */);
}
```

The cause is that the internal response handler unconditionally calls `res.json()` when the response comes back. 

This is possible to handle with a custom `responseHandler` argument in the api configuration:

```js
deleteUser: builder.mutation<void, string>({
  query: (id) => ({
    url: `/users/${id}`,
    method: 'DELETE',
    responseHandler: async () => null,
  }),
  invalidates: [USERS],
}),
```

So as-is, it is possible to handle this situation. But, I was wondering what you all thought about adding some default support for this case? I'm not sure how common of a use-case this is, but it does seem like it would be good to not reject by default.

If you like the idea, I'd love to take a crack at making the change myself.

phryneas:
A shorter version that is currently is possible is `responseHandler: 'text'`.

But yeah, a PR that automatically handles the edge case of empty-string responses as empty string instead of trying to parse it as JSON would be welcome I guess - what do you think @msutkowski ?

msutkowski:
Sounds reasonable to me!

phryneas:
Just merged #91, thank you for bringing this to our attention & the PR :+1:

PR #91, opened 2020-12-05
Handle empty json response

airjp73:
Addresses #89.

Of course, I'm happy to fix anything if it's not quite right 👍 .

### DIFF
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -2,17 +2,34 @@ import { joinUrls } from './utils';
 import { isPlainObject } from '@reduxjs/toolkit';
 import { BaseQueryFn } from './apiTypes';
 
+export type ResponseHandler = 'json' | 'text' | ((response: Response) => Promise<any>);
+
 export interface FetchArgs extends RequestInit {
   url: string;
   params?: Record<string, any>;
   body?: any;
-  responseHandler?: 'json' | 'text' | ((response: Response) => Promise<any>);
+  responseHandler?: ResponseHandler;
   validateStatus?: (response: Response, body: any) => boolean;
 }
 
 const defaultValidateStatus = (response: Response) => response.status >= 200 && response.status <= 299;
 
 const isJsonContentType = (headers: Headers) => headers.get('content-type')?.trim()?.startsWith('application/json');
+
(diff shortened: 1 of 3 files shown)


src/fetchBaseQuery.ts
```diff
@@ -2,18 +2,35 @@ import { joinUrls } from './utils';
...
+    return responseHandler(response);
+  }
+
+  if (responseHandler === 'text') {
+    return response.text();
+  }
+
+  if (responseHandler === 'json') {
+    const text = await response.text();
+    return text.length ? JSON.parse(text) : null;
```
phryneas, line 30: Hmm.
Not sure what this should be.
I'm tending more towards `undefined`.
@msutkowski @markerikson can you take a look?

airjp73: Yeah, I can definitely see the case for this being `undefined` instead. Happy to update once the others weigh in.

msutkowski: undefined is probably better... the body was technically undefined in this case, ya?

test/fetchBaseQuery.test.tsx
```diff
@@ -89,6 +89,22 @@ describe('fetchBaseQuery', () => {
       expect(res.data).toEqual({ value: 'success' });
     });
 
+    it('should return null for a simple GET request when the response is null', async () => {
```
phryneas, line 92:
```suggestion
    it('should return undefined for a simple GET request when the response is empty', async () => {
```

test/mocks/server.ts
```diff
@@ -7,6 +7,7 @@ import { rest } from 'msw';
 export const server = setupServer(
   rest.get('http://example.com/success', (_, res, ctx) => res(ctx.json({ value: 'success' }))),
   rest.post('http://example.com/success', (_, res, ctx) => res(ctx.json({ value: 'success' }))),
+  rest.get('http://example.com/empty', (_, res, ctx) => res(ctx.json(null))),
```
phryneas, line 10:
```suggestion
  rest.get('http://example.com/empty', (_, res, ctx) => res(ctx.body(''))),
```

phryneas:
Let's get this merged.

PR #93, opened 2020-12-07
Major WIP: experiment with using reducerPath and unscoping actions

msutkowski:
This is just an experiment. I was trying to pass through `reducerPath` so that we only needed one middleware, but there is a ton of stuff that is intermingled. 

I'm not entirely sure if it's worth the effort to do this. We'd need to find a way to get the endpointDefinitions to a `buildMiddleware` function as well as `buildThunks` without having them be scoped to a specific API?


phryneas:
Good point, I didn't think about **that** :facepalm:

Considering the complexity from that point on, it might actually not be worth it. :disappointed:

phryneas:
I guess I'll close this for now - we can always start another attempt at a later time

PR #100, opened 2020-12-10
remove phantom type identifying state

phryneas:


phryneas:
to be merged shortly after #90

PR #102, opened 2020-12-11
Enhance endpoints

phryneas:
This would be used to enhance auto-generated endpoints in a second file under manual source control. 
So far: no ability to change any types, just the possibility to override a query or add a `provides`/`invalidates` or something like that. That *should* do it for most use cases though.

Thoughts?


kahirokunn:
Looks pretty good to me 👍

kahirokunn:
I think Enhance endpoints is a great feature to encourage authors of third party libraries to contribute features.

msutkowski:
This looks good for endpoints. The only issue I see currently is the baseQuery. If the goal is to have an autogenerated file, then override it, you'd also want to be able to do the same with baseQuery to add additional enhancers.

phryneas:
Yeah, I've also noticed that we need to override `entityTypes`, because otherwise you can't add `invalidates`/`provides` here. I'll extend that a bit.

msutkowski (approved):

PR #103, opened 2020-12-11
docs: update comparison

tannerlinsley:
Adds "declarative" as an option for API definition for React Query. With v3, you can now configure all of your queries and mutations up front before calling them.


phryneas:
Ah, nice. Gotta check this out. Just out of interest: did we inspire you or was that on your list anyways? ^^

phryneas (approved):

tannerlinsley:
We've been working on it almost since v2 shipped earlier this year. But to be fair, it's still not the "generative" declarations like RTK. For example, RTK does a lot of cool heavy-lifting for you via configutation and entity types and auto-invalidation etc, (super cool btw), where RQ, despite supporting a declarative api, still requires that you wire up queries/mutations to work together.  This does bring RQ one step closer to being able to do the auto-link stuff too, but we'll have to see if we go in that direction with the core, or just ship that out to utility wrappers / plugins/ who knows :)

phryneas:
To be fair, while it's cool, it also adds a bit of a first hurdle of concepts to learn, so it's maybe better if we don't all have the same featureset in the end :sweat_smile: 

I'm just skimming the docs over at https://react-query-beta.tanstack.com/overview - I assume I should be looking at the right place, but can't find anything regarding a declarative definition. Can you point me to the right spot?

Oh, and I noticed that you are adding the size of RTK and RTK-query in the size comparison table. That's not 100% correct as in that case you're adding `immer` twice. So you can substract another 5.4kb from that ;)

But it's *very* cool to be on that list :tada: 
Although it's a bit sad to see urql not on that list any more - it's a cool library and was a big inspiration for the `entityTypes` (it's their default cache invalidation model). But yeah, space is limited I guess.

msutkowski (approved):
Thanks @tannerlinsley! I think you already mentioned this, but we should probably just pull the react-query table directly? We're going to do another release here shortly and should probably just update both at the same time.

phryneas:
Happy release day!

Issue #88, opened 2020-12-05
Investigate SSR for RTK Query


markerikson:
We had a couple questions about "will RTK Query support SSR?"

None of us maintainers have any real experience with SSR.  So, requests for more details from the community:

- How do you currently handle SSR work with vanilla Redux / RTK?
- What's the data flow sequence like there?
- How does SSR work with non-Redux options like React-Query, SWR, Urql, or Apollo?
- What use cases would you like to have working here?

Possibly related docs/issues:

- https://github.com/vercel/swr/issues/11
- https://dev.to/able/prefetching-data-in-a-next-js-application-with-swr-c63
- https://react-query.tanstack.com/docs/guides/ssr

JNaftali:
When I've done SSR with Redux in the past the basic flow has been
* dispatch actions on the server (awaiting async thunk actions as necessary: not sure how that would work with other middlewares)
* render your react app (rendering on the server is single pass)
* serialize the state of the redux store as JSON (sent to the client serialized in the dom or via api call)
* use that json as the initial state of the redux store when rendering on the client

There are 2 basic approaches for how the libs you mentioned support SSR.
1. Fetch initial data outside of React then pass it into the tree via props
2. Multiple single-pass renders. The first render kicks off fetching (which is then stored in some external data cache outside of the react tree). Once fetches have resolved a second render happens. This time the data is in the cache so your app renders fully.

The first approach has the advantage of being relatively simple to implement and reason about, but because you have to fetch the initial data yourself it can be difficult to know what data deeply nested components need. On the other hand, multiple renders can be very tricky to implement (especially if you're working within the confines of something like Nextjs). Eventually _**Suspense for Data Fetching™**_ will give us the utility of the second approach with the simplicity of the first and we can all stop talking about this and take naps instead.

Of the libs you mentioned, only SWR doesn't have any guide or support for the second approach. All support the first approach (React-Query and SWR hooks accept initial data, Urql and Apollo caches can be populated before being provided in the react tree).  https://github.com/FormidableLabs/react-ssr-prepass might be a useful resource for y'all - it's what urql uses to do the multiple renders thing, but has no urql-specific bits in it.

It looks like the first approach should be possible with RTK Query now - just `await dispatch(api.getWhatever.initialize())` before rendering the react app on the server. I can try to whip up a simple nextjs example of that to see if I'm completely off-base

JNaftali:
[...]

PR #105, opened 2020-12-12
Allow for a custom fetch fn, fix isOnline

msutkowski:
Addresses points brought up in #88. I'll work on adding docs + SSR-related content as we navigate through it.


phryneas (approved):
Looks good :+1:

Issue #96, opened 2020-12-08
Using selectors with query data


cole-robertson:
Is it possible to use selectors with the data retrieved by rtk-query just like you can with normal RTK state? 

There are scenarios where I want to make use of the composed selectors functionality of reselect, I was wondering if I can do this with rtk-query data. I also have the pattern of creating a thunk that will call my api to update my data and then replace that data point within my reducer. Is there a way to use thunks in this fashion with rtk-query?

Issue #84, opened 2020-12-02
Syncing slices of redux state with subquery results


talolard:
Relates to #79 

I [commented on twitter](https://twitter.com/de_stroy/status/1334191220927590400) about how integrating RQ with redux hurts. 
I'll describe what's hurting for us and add a video below, I have problems but no good solutions to offer: 

We have a data model and API calls that look like this
```ts
export interface X {
    id:string
    data:Record<any, any>
}
interface XSummary {
    xId:string
    stat1:number,
    stat2:number
    
}
export interface RelatesToX {
    id:string
    xId:string //The id of the X
}
function getAllXSummary():Array<XSummary>
/* Expensive to call because it goes over all the X */ 

function getOneXSummary(xId:string) :XSummary

function getXWithRelated() :{x:X,related:Array<RelatesToX>} 

function changeARelatedToX(relatedToXId:string,something:any)
```

We display the list of XSummary to the user and when they click on one we should them the X with all of the RelatedToX.

When the user calls ***changeARelatedToX*** we'd like to update the single ***XSummary*** in the list, without calling **getAllXSummary** because it's expensive. 

Ideally, when the user triggers a call to ***changeARelatedToX*** it would subsequently call ***getOneXSummary*** for the given xId. 

# Why It's Hard with React-Query

React Query caches the the list of XSummary somewhere magical, so it's not clear what the result of ***getOneXSummary*** should update. Once the call to ***getAllXSummary*** is complete, we want to keep track of the individual XSummary states so that we can make the user happy (video below), but it's just not clear how to do that

# Real World User Case
I made a loom of where we are struggling. 
https://www.loom.com/share/91ce26804c8740e1bc1f741f067a9b9b

# Towards a solution

This seems like a variation on #79. I want my list query to register it's results in a clearly indicated state slice, and be able to associate a query with each item in that slice, where the originating query to populate the slice is the query that brought back the entire list. 

Then, I'd also like a canonical way to trigger the "getter" for a particular item in the slice programmatically from some seemingly unrelated query. 

# Why that's hard in RQ
Possibly repeating myself, but in react-query I don't have a clear way to 
1. associate an item that came back from a "list query" with it's own query. e.g. what is the query key for that item ? 
2. Associate each item in a (displayed) list with it's own query that is pre-propulated with the results of the original list query.

phryneas:
So, from the back of my head, right now with the existing APIs, you would have two options:

1. modify the list in "the cache"
or
[...]

PR #106, opened 2020-12-12
split query hooks, add sub-selector

### LINKED ISSUE #79: Investigate or document data selection patterns
Seen a couple comments on this idea and I think there's good reason to support it.

Right now, it's possible to have both a "fetch all posts" query, and a "fetch one post by ID" query.  But, those end up as two separate queries with separate caches.  It seems like there's a missed use case: "fetch all the posts", but then "select one item out of that 'all posts' cache by ID".  Loosely put, I'm picturing something like:

```js
  const selectPostById = createSelector(
    postsApi.selectors.selectAllPosts,
    (state, postId) => postId,
    (posts, postId) => posts.find(p => p.id === postId)
  )

// later
  const post = useSelector(state => selectPostById(state, props.postId))
```

Maybe that's entirely doable already and I just haven't actually looked at the API enough to know if this is correct :) 

Related: it's a common Redux pattern to have a list parent select _just_ an array of item IDs from the store, as a stable reference, and then render a list of `<ListItem key={id} id={id} />`, possibly with the children being memoized.  Then, if a data item does get updated, the parent either continues to get the same array of IDs as doesn't re-render at all, or at least all the children are memoized and don't re-render.

How feasible is it to select item IDs from this, ideally using a stable array ref?

phryneas:
This splits `useQuery` up into `useQueryState` and `useQuerySubscription`.

`useQueryState` returns the enhanced querySubState like before, but allows for an option `subSelector` to select only a partial result.

`useQuerySubscription` starts a query, subscribes to it and only returns `{refetch}`.

`useQuery` calls both of them. Right now, `useQuery` does not accept a subselector yet, but I might change that, depending on the resulting complexity.

closes #79  #84 #96

### DIFF
--- a/examples/react/src/features/posts/PostsManager.tsx
+++ b/examples/react/src/features/posts/PostsManager.tsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+import React, { useCallback, useState } from 'react';
+import { useEffect } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Route, Switch, useHistory } from 'react-router-dom';
 import {
@@ -7,6 +8,7 @@ import {
   useGetPostsQuery,
   useLoginMutation,
   useGetErrorProneQuery,
+  postApi,
 } from '../../app/services/posts';
 import { selectIsAuthenticated, logout } from '../auth/authSlice';
 import { PostDetail } from './PostDetail';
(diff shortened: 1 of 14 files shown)


msutkowski (commented):
This looks great so far. It would be excellent if we could get the `subSelector` into `useQuery` itself so there is only ever one 'entry point' for usage. If it's not feasible, I don't think the one-off selector usage is currently bad, but I could see some frustrations with it. I'm just not sure how common of a use case it would be 🤔 , but as Mark has pointed out, it's most likely going to be a demand from redux-specific users.

```js
const PostFromSelector = () => {
  const post = postApi.endpoints.getPosts.useQueryState(undefined, {
    subSelector: ({ data }) => data?.find((entry) => entry.id === 6),
  });

  return (
    <div>
      {post ? (
        <div>
          <h3>From a selector!</h3>
          {JSON.stringify(post)}
        </div>
      ) : null}
    </div>
  );
};
```

markerikson:
Not thrilled with the name `subSelector`. 

Perhaps `selectFromResult`, or just `select`?

msutkowski (approved):

phryneas:
I'll merge this now, we'll rename it in the docs PR Matt is gonna do later.

Issue #107, opened 2020-12-13
fetchBaseQuery: Allow to pass `undefined` as value into headers


phryneas:
See this generated query from the OpenAPI petStore example:

https://github.com/rtk-incubator/rtk-query-codegen/blob/main/test/generated.ts#L24-L26
```ts
        deletePet: build.mutation<DeletePetResponse, DeletePetQueryArg>({
            query: queryArg => ({ url: `/pet/${queryArg.petId}`, headers: { api_key: queryArg.apiKey } })
        }),
```

since `api_key` is optional in `DeletePetQueryArg`, this line currently gives an error. 

We should allow for `undefined` in headers & filter these out.

PR #108, opened 2020-12-13
Allow more flexible headers that can be undefined

msutkowski:
Addresses #107. I couldn't override things while extending the interfaces. Let me know if there is a better way to do this :)

### DIFF
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -1,78 +1,14 @@
 import { joinUrls } from './utils';
 import { isPlainObject } from '@reduxjs/toolkit';
 import { BaseQueryFn } from './apiTypes';
+import { Override } from './tsHelpers';
 
 export type ResponseHandler = 'json' | 'text' | ((response: Response) => Promise<any>);
 
-// We need to control most of these types to allow for more flexible headers
-interface Headers {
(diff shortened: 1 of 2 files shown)


src/fetchBaseQuery.ts
```diff
@@ -69,13 +139,19 @@ export function fetchBaseQuery({
...
+    });
+
+    if (!tempHeaders.has('content-type')) {
+      tempHeaders.set('content-type', 'application/json');
     }
 
-    if (body && isPlainObject(body) && isJsonContentType(config.headers)) {
+    if (body && isPlainObject(body) && isJsonContentType(tempHeaders)) {
       config.body = JSON.stringify(body);
     }
```
msutkowski, line 157: None of this really seems like a good idea to me. Instead of passing instances of Headers around, we might want to go the redaxios way and just use objects as it'd be much safer than this. 🤷 I suppose we could also check for `instanceof Headers`, and if false, delete the undefined keys before converting them.

phryneas: how about this?

msutkowski: Your approach looks right :). Those dang TS overrides.... learning something new every day here 🎉. I'm going to check the test cases for all of this, and then merge.

PR #116, opened 2020-12-16
split up into "core" & "redux-hooks" module

phryneas:


src/react-hooks/index.ts
```diff
@@ -0,0 +1,7 @@
+import { coreModule } from '../core/module';
+import { buildCreateApi } from '../createApi';
+import { reactHooksModule } from './module';
+
+const createApi = buildCreateApi(coreModule, reactHooksModule);
+
+export { createApi, reactHooksModule };
```
phryneas, line 7: Just looking at this part makes it kinda worth it ^^

msutkowski: Agreed. This is what caught me.

msutkowski (approved):

PR #123, opened 2020-12-18
Make prepareHeaders support async functions

schadenn:
I'm trying to use the `prepareHeaders` function to add an `accessToken` to my API requests. Unfortunately the function I'd have to use to get the `accessToken` in our setup is asynchronous.

So my proposal would be: Make the `prepareHeaders` function support async functions.
Since this is being called in an asynchronous function anyways there's not a lot to change and it'd save us a lot of overhead by not having to put the `accessToken` into state first and then get it from there.

Hopefully it'll also help some other users :)

### DIFF
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -1,7 +1,7 @@
 import { joinUrls } from './utils';
 import { isPlainObject } from '@reduxjs/toolkit';
 import { BaseQueryFn } from './baseQueryTypes';
-import { Override } from './tsHelpers';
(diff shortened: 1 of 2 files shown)


msutkowski:
@schadenn Thanks for the PR! Do you mind if I add a quick test on top of this or would you be interested in doing that?

schadenn:
I'll add it.

msutkowski:
> I'll add it.

Thank you! 🥇

schadenn:
Done :)

msutkowski (approved):

PR #127, opened 2020-12-23
Link directly to #redux on Reactiflux

nickmccurdy:
I think you'll want to also `git cherry-pick` this into `next` so the documentation updates are reflected both in the readme and on the website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 <p align="center">
   <img src="https://img.shields.io/badge/ALPHA-pre%20release-orange">
-   <a href="https://discord.gg/reactiflux" target="_blank">
+   <a href="https://discord.gg/0ZcbPKXt5bZ6au5t" target="_blank">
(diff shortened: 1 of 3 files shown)


msutkowski (approved):

msutkowski:
I think we'll leave this as the question in the RTK docs is pending - will this direct link work with the need to opt-in to the server ToS or whatnot? https://github.com/reduxjs/redux-toolkit/pull/855#issuecomment-750841560

nickmccurdy:
> will this direct link work with the need to opt-in to the server ToS or whatnot? [reduxjs/redux-toolkit#855 (comment)](https://github.com/reduxjs/redux-toolkit/pull/855#issuecomment-750841560)

Yes. See https://github.com/reduxjs/redux-toolkit/pull/855#issuecomment-751215457.

phryneas (approved):
Then let's get this in :)

Issue #85, opened 2020-12-03
Naming of hook option for refetch on mount/arg change.


phryneas:
So, we have a functionality`in #80, but are missing a good name and are hoping for some feedback from the community.

Imagine these three use cases:

### basic/no option
```ts
// This will only fetch data if data for useGetPostsQuery({ page: 1 }) has not been fetched yet by this or another component.
// Similarly, changing `page` to `2`, it will only fetch data if no data is present yet.
const result = useGetPostsQuery({ page: 1 })
```

### always refresh/refetch
```ts
// This will fetch data on first mount and on every change of `page`
const result = useGetPostsQuery({ page: 1 }, { XXXXonYYYY: true })
```

### refresh/refetch if data is older than X seconds
```ts
// This will fetch data on first mount and change of `page`, but only if there is no data in the cache or if that data is older than 20 seconds.
const result = useGetPostsQuery({ page: 1 }, { XXXXonYYYY: 20 })
```

So, now the question is: What is a correct name for `XXXXonYYYY`?

`refetchOnMount` does not seem quite right, because this is also a `refetchOnArgChange`. `refetchOnMountOrArgChange` is horribly long, albeit very on-spot.

So far, we've come up with `refetchOnSubscribe`, `refetchOnUse`, `refetchOnQuery`, `refreshOnSubscribe`, `refreshOnUse` and `refreshOnQuery`, but we don't really have a strong preference on either of these.

So... maybe we can get some ideas or preferences from the community on this one?

michaelaflores:
Is `refetchOnRender` accurate?

adokce:
I understand that long names would usually drive people away from using the api, but `refetchOnMountOrArgChange` sounds beautiful (and perfectly descriptive) to me

meesvandongen:
Simply "refetch" may be enough.

phryneas:
> Is `refetchOnRender` accurate?

No, if you rerender 20 times, but the argument does not change, it will not even evaluate refetching. It's only mount & actual change of argument (an object as argument will be shallow-compared with the last render).

Shrugsy:
> Is `refetchOnRender` accurate?

I would read that as re-fetching *every* render, including renders after the initial one when already fetched, as well as when args did not change (which would cause an infinite loop)

phryneas:
> Simply "refetch" may be enough.

Interesting idea :)
We already have that as an active method returned from the hook, so that might be a bit confusing in the docs :thinking:

[...]

Issue #52, opened 2020-11-25
Enhancing Endpoint definitions.


phryneas:
Just an idea: where `injectEndpoints` creates/overrides endpoints, it might be a good idea to have a `enhanceEndpoints` helper.

That helper would only allow to adjust existing endpoints, e.g. by adding `provides`/`invalidates`.

That way one file could contain a "generated from OpenAPI" api definition and you could chain a second file behind that, that would only add certain properties and do light adjustments - allowing for continuous re-creation of the generated file.

phryneas:
I've contacted the maintainers of `oazapfts` as a possible client generator over at https://github.com/cellular/oazapfts/issues/41

markerikson:
Oh hey, that looks interesting.

I tried out the official `openapi-generator` Axios+TS thing last year, and was _not_ impressed with its output.  5000+ lines of boilerplate TS for just a few endpoints.

phryneas:
current progress of code generation: 
![image](https://user-images.githubusercontent.com/4282439/101289589-bb168180-37fd-11eb-9104-fcae4b52fa07.png)

https://github.com/phryneas/rtk-query-codegen-openapi/blob/main/src/generate.ts
If you want to play around with it:
requires this branch from 
https://github.com/phryneas/oazapfts/tree/expose-class
yalc it in & run the generate.ts with ts-node

kahirokunn:
@phryneas It works for me ❤️ 
The code also uses the TypeScript API, which I thought was awesome! 👍

kahirokunn:
@phryneas 

What is current progress?
Is there anything I can do to help? 👍  (eg. cli interface

kahirokunn:
For the first stage, unoptimized code should be sufficient.

phryneas:
Hey @kahirokunn,
I'm sick at the moment so I can only work on stuff for like one hour or so and have to rest then, so this has been quite delayed.
The actual state now produces actually working code - see https://github.com/phryneas/rtk-query-codegen-openapi/blob/main/test/generated.ts

If you could help out there, that would be great. I would try & do some work on the rtk-query side then.
Stuff currently outstanding:

* add a cli
* allow to specify some overrides for example for baseUrl
* generation of provides/invalidates is completely untouched now, so experimentation in that direction is very welcome

kahirokunn:
@phryneas Oh, sorry 🙏 Take care 😢 

I first created a pull request for the first "add cli".
[...]

PR #130, opened 2020-12-24
add onStart onError onSuccess hooks to query

kahirokunn:

PR #135, opened 2020-12-26
Prepare 0.2.0

### LINKED ISSUE #114: Even if the API returns a 404, Promise is still resolved.
https://rtk-query-docs.netlify.app/concepts/mutations#basic-mutation
So never come catch section 😢 

The resolved values look something like this.

スクリーンショット 2020-12-16 13 15 51

phryneas:
Closes #114 #96 #52 #85


phryneas:
Today we're releasing version 0.2.0 - the second alpha - of RTK-Query. So, what's new?

We've had some discussion & requests around the possibility to just subscribe to part of a query result instead of the full result and `isFetching`/`isSuccess`/etc., to have more granluar rerenders.
This is now possible with the `selectFromResult` option to `useQuery`.

One thing that we had to add to make this possible is "structural sharing". 
So, if you have a second request that comes back with (partially) similar data, we try to recycle as many object references from the last query result as possible, to prevent unneccessary rerender.

If you want to only select query result data or only trigger a query without caring for any result, you can now use the `useQueryState` and `useQuerySubscription` hooks. (https://github.com/rtk-incubator/rtk-query/pull/106)




Another big thing is code generation. You can now generate full API definitions from OpenAPI definitions. Big thanks to @fgnass who let's us use the type generation from his `oazapfts` library.

If you want to give it a try, run `curl -o petstore.json https://petstore3.swagger.io/api/v3/openapi.json;
npx @rtk-incubator/rtk-query-codegen-openapi petstore.json > petstore-api.generated.ts`.
Docs are over at https://rtk-query-docs.netlify.app/concepts/code-generation



The third big thing is a new build configuration to enable tree-shaking. Bundle sizes are significantly down due to that. Also, RTK-Query has been modularized, so you can now use `createPureApi` if you don't need the React hooks

The module system also allows you to write your own `createApi` plugins. See https://rtk-query-docs.netlify.app/concepts/customizing-create-api


In addition, we've added some new options to tweak the refetch-behaviour of the hooks.
There now are `refetchOnFocus` and `refetchOnReconnect` options for the hooks. (https://github.com/rtk-incubator/rtk-query/pull/90)

Also, we now have a `refetchOnMountOrArgChange` option to trigger a refetch when a component is mounted/changes it's query argument even if there is already data in the cache for the target route. You can specify a "minumum age" here. (https://github.com/rtk-incubator/rtk-query/pull/80)



Also, some smaller changes:

* We had a bug report & fix PR from https://twitter.com/PettengillAaron for the edge case when an api returned an empty response body. Parsing that as JSON was probably not a good idea. Thanks! (https://github.com/rtk-incubator/rtk-query/pull/91)

* You can now pass in a custom `fetch` function into `fetchBaseQuery` - this came up in a discussion on server side rendering as a way to make better use of `fetch` ponyfills. Thanks to https://twitter.com/JoshMarantz for bringing this up. (https://github.com/rtk-incubator/rtk-query/pull/105)

* You can now call `.unwrap` on the result from the `useMutation` hook and `mutation.initiate` action creator to have the promise resolve with the returned value or throw an exception. If people like this API, we might add this to RTKs `createAsncThunk` as an alternative to `unwrapResult`. Thanks for https://twitter.com/kahirokunn for bringing this up. (https://github.com/rtk-incubator/rtk-query/pull/117)
Also, be aware: this changed the general return value of the result from the `useMutation` hook from the resulting state to `{data: ...} | {error: ...}` in case you're not unwrapping.

* `prepareHeader` can now be an async function. Thanks to https://github.com/schadenn for the contribution. (https://github.com/rtk-incubator/rtk-query/pull/123)

* Query endpoint definitions now support the same `onStart`, `onError` and `onSuccess` options that mutation endpoints already did. Thanks to https://twitter.com/kahirokunn. (https://github.com/rtk-incubator/rtk-query/pull/130)

* `api.internalActions.prefetchThunk` was renamed to `api.util.prefetchThunk` (https://github.com/rtk-incubator/rtk-query/pull/133)

A big thanks to everyone I forgot to mention here who discussed with us on github, in @reactiflux, or here on Twitter.
And an enormous "thank you" to https://twitter.com/de_stroy who did an incredible amount of PRs, Code Reviews and Documentation.

msutkowski (approved):

PR #140, opened 2021-01-01
Support FormData in fetchBaseQuery

msutkowski:
- Does not set `application/json` content-type if the body is type of FormData.
- Removes a header check from tests, as we won't set application/json unless the body is an object


test/fetchBaseQuery.test.tsx
```diff
@@ -462,3 +457,23 @@ describe('fetchFn', () => {
...
+      { url: '/echo', method: 'POST', body },
+      {
+        signal: undefined,
+        dispatch: storeRef.store.dispatch,
+        getState: storeRef.store.getState,
+      },
+      {}
+    ));
+    expect(request.headers['content-type']).not.toContain('application/json');
+  });
```
msutkowski, line 478: The browser will specify the content-type of multipart/form-data with the correct boundary. Couldn't find a way to test this specifically outside of a browser, but this should be sufficient.

phryneas (approved):
Looks good to me.

Issue #138, opened 2021-01-01
rtk-query requires react type definitions (not framework-agnostic)


lacolaco:
I've tried to use RTK-Query in an Angular/Redux-Toolkit project but it requires type definitions from react and react-redux, which seems to be for Hooks support.

```
    node_modules/@rtk-incubator/rtk-query/dist/ts/react-hooks/module.d.ts:21:30 - error TS2307: Cannot find module 'react-redux'.
    
    21     useStore?: typeof import('react-redux').useStore;
                                    ~~~~~~~~~~~~~
    node_modules/@rtk-incubator/rtk-query/dist/ts/react-hooks/ApiProvider.d.ts:1:25 - error TS2307: Cannot find module 'react'.
    
    1 import { Context } from 'react';
                              ~~~~~~~
    node_modules/@rtk-incubator/rtk-query/dist/ts/react-hooks/ApiProvider.d.ts:2:40 - error TS2307: Cannot find module 'react-redux'.
    
    2 import { ReactReduxContextValue } from 'react-redux';
                                             ~~~~~~~~~~~~~
    node_modules/@rtk-incubator/rtk-query/dist/ts/react-hooks/ApiProvider.d.ts:16:5 - error TS2503: Cannot find namespace 'JSX'.
    
    16 }): JSX.Element;
           ~~~
```

Is this by design?

![image](https://user-images.githubusercontent.com/1529180/103431695-af10ba80-4c17-11eb-8d55-3180b61c4681.png)

markerikson:
Nope.  This is, as clearly labeled, still an experimental alpha, and thus subject to both bugs and unstable changes :)  @phryneas just got done trying to rearrange one of the core APIs to be more UI-agnostic, but it's entirely possible something got missed.

Can you post a project that reproduces the issue?

lacolaco:
Sure. Here it is. https://github.com/lacolaco/rtkquery-playground

1. `yarn install`
2. `yarn build`
3. Errors!

Wishing you a Happy New Year!

phryneas:
Could you try it with the platform-agnostic `createBaseApi` entrypoint? `createApi` contains the react-redux hooks. We might rename that to `createReactHooksApi` but I didn't want to introduce any breaking changes between alpha 1 & 2.

phryneas:
Hmm. Just tried it. At runtime, this would do it.
[...]

PR #141, opened 2021-01-01
extra `/react` entrypoint

phryneas:
let's see how this works out

Closes #141

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,20 +8,21 @@
   "main": "dist/index.js",
   "module": "dist/esm/index.js",
   "sideEffects": false,
-  "types": "dist/ts/index.d.ts",
+  "types": "dist/esm/ts/index.d.ts",
   "typesVersions": {
     ">=4.1": {
       "dist": [
-        "dist/ts-4.1/index.d.ts"
+        "dist/esm/ts-4.1/index.d.ts"
       ],
-      "dist/ts/*": [
-        "dist/ts-4.1/*"
+      "dist/esm/ts/*": [
+        "dist/esm/ts-4.1/*"
       ]
     }
   },
   "files": [
     "dist",
-    "src"
+    "src",
+    "react"
   ],
   "publishConfig": {
     "access": "public"
@@ -31,7 +32,7 @@
   },
   "scripts": {
     "start": "tsdx watch",
-    "build": "rollup -c && node post-build.js",
+    "build": "rm -r dist; rollup -c && node post-build.js",
     "test": "tsdx test",
     "lint": "tsdx lint",
     "prepare": "yarn build",
@@ -67,13 +68,18 @@
       "limit": "15 KB"
     },
     {
-      "name": "createBaseApi + setupListeners",
-      "path": "dist/esm/index.js",
-      "import": "{ createBaseApi, setupListeners }"
+      "name": "ESM full (React)",
+      "path": "dist/esm/react.js",
+      "limit": "15 KB"
     },
     {
       "name": "createApi + setupListeners",
       "path": "dist/esm/index.js",
+      "import": "{ createApi, setupListeners }"
+    },
+    {
+      "name": "createApi (React) + setupListeners",
+      "path": "dist/esm/react.js",
       "import": "{ createApi, setupListeners }"
     },
     {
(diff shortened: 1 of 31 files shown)


phryneas:
So, I *think* this should cover everything.

@Andarist, I would be very grateful if you could take a look at this if I did anything horribly wrong :)

rollup.config.js
```diff
@@ -5,10 +5,11 @@ import replace from '@rollup/plugin-replace';
 
 /** @type {import("rollup").RollupOptions} */
 const defaultConfig = {
-  input: 'src/index.ts',
+  input: ['src/index.ts', 'src/react.ts'],
```
Andarist, line 8: q: any special reason why you'd like to keep using `preserveModules`?

Andarist:
@phryneas nothing obviously broken here 👍 it's a lot of delicate configs so I can't give any guarantees that this works, but I know you have done some manual testing already, so 🤞

rollup.config.js
phryneas: Compared to a single-file ESM build, tree shaking seems to be working a LOT better.

With `preserveModules` (0.2.0):
image
Normal rollup build (0.1.0):
image

phryneas:
@Andarist one bonus question: do you know how to keep it from stripping `// @ts-ignore` comments?
The example from #138 still type-checks even non-included files & errors as soon as it encounters the `react-redux` type import. `/* @ts-ignore */` is being preserved through build, but TS does not accept that.

rollup.config.js
Andarist: Was expecting this answer and it's a good one. Just a note for completeness - this doesn't quite improve tree-shaking (given my own definition of what tree-shaking is). It allows bundlers to leverage `"sideEffects": false` - but this flag is a hack around tree-shaking as ideal tree-shaking in practice is sometimes hard to achieve (due to initialization side-effects that include function calls and property accesses). 

Both things are not mutually exclusive though as `"sideEffects": false` relies heavily on the file boundaries that you have set up as an author which might not produce 100% ideal results either.

msutkowski (commented):
This looks good to me. We'll need to update docs as well before a release. I only have one real question - do the types need to be moved from `/ts/` -> `/esm/ts` for a particular reason? 

I'm wondering if this should also be documented in release notes, as it could be a breaking change for some folks?

phryneas:
> This looks good to me. We'll need to update docs as well before a release. I only have one real question - do the types need to be moved from `/ts/` -> `/esm/ts` for a particular reason?

Simplicity of the build. We'd need an additional step otherwise, as only the ESM build builds all entry points, but TS wants to emit only to sub-folders of the ESM build.
 
> I'm wondering if this should also be documented in release notes, as it could be a breaking change for some folks?

Hmm. It's not really "documented API" in my mind. Only what is exported from index (or other "official entry points") has a guarantee to stay the same between versions, and that entry point still exports the same types as before (minus the react ones).

package.json
```diff
@@ -8,20 +8,21 @@
...
   "typesVersions": {
     ">=4.1": {
       "dist": [
-        "dist/ts-4.1/index.d.ts"
+        "dist/esm/ts-4.1/index.d.ts"
       ],
-      "dist/ts/*": [
-        "dist/ts-4.1/*"
+      "dist/esm/ts/*": [
+        "dist/esm/ts-4.1/*"
```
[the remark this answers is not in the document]
phryneas: Yup, before types were generated as part of the CJS build, now they're part of the ESM build. So they have to be in a subfolder of the ESM build for "rollup reasons":tm:.
And we can't move them back up because then the sourcemaps would not point to the correct relative paths any more.

msutkowski:
> > This looks good to me. We'll need to update docs as well before a release. I only have one real question - do the types need to be moved from `/ts/` -> `/esm/ts` for a particular reason?
> 
> Simplicity of the build. We'd need an additional step otherwise, as only the ESM build builds all entry points, but TS wants to emit only to sub-folders of the ESM build.
> 
> > I'm wondering if this should also be documented in release notes, as it could be a breaking change for some folks?
> 
> Hmm. It's not really "documented API" in my mind. Only what is exported from index (or other "official entry points") has a guarantee to stay the same between versions, and that entry point still exports the same types as before (minus the react ones).

Makes sense to me, thanks!

package.json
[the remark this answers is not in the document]
msutkowski: Thanks for the clarification!

msutkowski (approved):

phryneas:
Umh, yeah, forgot about this. LGTM.

PR #144, opened 2021-01-05
fix getState type

kahirokunn:


phryneas:
Hmm, this would add a dependency to `react-redux` from the (see #141) framework-agnostic entry point `/`, causing build failures for many people.

So we won't be able to add that one, I'm sorry. Guess it'll have to be a manual cast for now.


All that aside: 

`DefaultRootState` is not anything really official - that's just something someone at DefinitelyTyped added to the types.  
I mean, the pattern itself is a cool idea, but it's also pretty dangerous as just about any other module can now accidentally "poison" your global state structure just by being installed as a dependency.  
We're still not sure if we want to embrace that pattern of module augmentation or not - in any case it's not part of any official documentation or types of any officially maintained redux module.

Kinda the downside of having the types externally is that people always add new stuff without any real sense of ownership or communication. :/

kahirokunn:
I see.
Sorry, Thx for your explanation!

phryneas:
No need to be sorry, if we wouldn't have the need to keep `/` clean and `DefaultRootState` would be a bit more official, this would have been a great addition.

Thanks for that!

PR #152, opened 2021-01-18
SSR Support: delete substate.error property (Next.js getServerSideProps)

gfortaine:
Addresses points brought up in https://github.com/rtk-incubator/rtk-query/issues/88#issuecomment-762288424.

### DIFF
--- a/src/core/buildSlice.ts
+++ b/src/core/buildSlice.ts
@@ -98,7 +98,7 @@ export function buildSlice({
             if (substate.requestId !== meta.requestId) return;
             substate.status = QueryStatus.fulfilled;
             substate.data = copyWithStructuralSharing(substate.data, payload.result);
-            substate.error = undefined;
+            delete substate.error;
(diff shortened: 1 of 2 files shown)


phryneas:
Hmm. I'm honestly flabbergasted by the stuff next.js is pulling there, but oh well.

I'd suggest we do `delete substate.error` here though - not unsetting it at all would lead to pretty weird behaviour in some situations.

msutkowski:
How does everyone feel about just defaulting to `null` instead of `undefined`? Is that a better long-term solution for this?

phryneas:
Well, it could also just not be present yet or have never been set. And I kinda don't see the point of setting a property to null when it has never been accessed yet - especially since we would have to do that to others then as well consequently.
So I'd rather just use `delete X` instead of `X = undefined`.

gfortaine:
@phryneas PR updated with `delete substate.error` as suggested 👍

msutkowski (approved):

msutkowski:
Thanks again for this @gfortaine! 🎉

Issue #154, opened 2021-01-22
responseHandler should always be serializable (no function)


gfortaine:
It seems that Redux Toolkit complains when `responseHandler` [is a function](https://rtk-query-docs.netlify.app/api/fetchbasequery/#parsing-a-response) (because it is [not serializable](https://redux.js.org/style-guide/style-guide#do-not-put-non-serializable-values-in-state-or-actions)). Example repository :

https://github.com/gfortaine/rtk-query-nextjs-example/tree/responseHandler

PR #155, opened 2021-01-23
Redux Toolkit: responseHandler should always be serializable

gfortaine:
Addresses points brought up in https://github.com/rtk-incubator/rtk-query/issues/154

We have a use case where we need to consume the headers of the response. It looks like that we could handle this case through `responseHandler` by returning the raw Response : `(response) => response`. However, it seems that Redux Toolkit complains when `responseHandler` [is a function](https://rtk-query-docs.netlify.app/api/fetchbasequery/#parsing-a-response) (because it is [not serializable](https://redux.js.org/style-guide/style-guide#do-not-put-non-serializable-values-in-state-or-actions)). Thus, it appears that a suitable approach would be to simply add `'raw'` as a potential `responseHandler` type, then to handle the raw Response with `transformResponse` (see corresponding tests).

### DIFF
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -3,7 +3,7 @@ import { isPlainObject } from '@reduxjs/toolkit';
 import { BaseQueryFn } from './baseQueryTypes';
 import { MaybePromise, Override } from './tsHelpers';
 
-export type ResponseHandler = 'json' | 'text' | ((response: Response) => Promise<any>);
+export type ResponseHandler = 'json' | 'text' | 'raw';
 
 type CustomRequestInit = Override<
   RequestInit,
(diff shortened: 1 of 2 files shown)


phryneas:
I'll take a look tomorrow. It seems reasonable, but I want to check if we don't have other options that lead the the same pitfall.
You have seen that you can use the codesandbox build above meanwhile?

gfortaine:
OK, looking forward to your review tomorrow 👌By the way, many thanks for the codesandbox tip, I hadn’t seen it 👍

phryneas:
Unfortunately, there's a bunch of other options with the same problem. I've opened #156 to try and find a more broad solution. Can you take a look please?

gfortaine:
Solved by #156

PR #156, opened 2021-01-24
remove internalQueryArgs

phryneas:
This would address #154 and #155 in a more broad manner.

Also, I added `endpointDefinition` to `serializeQueryArgs` and renamed `endpoint` to `endpointDefinition` or `endpointName` everywhere to be more consistent.

Hope I didn't change a public API with it :sweat_smile:  (apart from `serializeQueryArgs`)

@msutkowski could you take a look at this? Also, @gfortaine, can you please test if this would solve your problem?

### DIFF
--- a/src/core/buildInitiate.ts
+++ b/src/core/buildInitiate.ts
@@ -86,27 +86,25 @@ export function buildInitiate<InternalQueryArgs>({
   const { unsubscribeQueryResult, unsubscribeMutationResult, updateSubscriptionOptions } = api.internalActions;
   return { buildInitiateQuery, buildInitiateMutation };
 
-  function buildInitiateQuery(endpoint: string, definition: QueryDefinition<any, any, any, any>) {
+  function buildInitiateQuery(endpointName: string, endpointDefinition: QueryDefinition<any, any, any, any>) {
     const queryAction: StartQueryActionCreator<any> = (
       arg,
       { subscribe = true, forceRefetch, subscriptionOptions } = {}
     ) => (dispatch, getState) => {
-      const internalQueryArgs = definition.query(arg);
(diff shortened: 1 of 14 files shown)


src/core/buildInitiate.ts
```diff
@@ -86,27 +86,25 @@ export function buildInitiate<InternalQueryArgs>({
...
 
-  function buildInitiateQuery(endpoint: string, definition: QueryDefinition<any, any, any, any>) {
+  function buildInitiateQuery(endpointName: string, endpointDefinition: QueryDefinition<any, any, any, any>) {
     const queryAction: StartQueryActionCreator<any> = (
       arg,
       { subscribe = true, forceRefetch, subscriptionOptions } = {}
     ) => (dispatch, getState) => {
-      const internalQueryArgs = definition.query(arg);
-      const queryCacheKey = serializeQueryArgs({ queryArgs: arg, internalQueryArgs, endpoint });
+      const queryCacheKey = serializeQueryArgs({ queryArgs: arg, endpointDefinition, endpointName: endpointName });
```
msutkowski, line 94:
```suggestion
      const queryCacheKey = serializeQueryArgs({ queryArgs: arg, endpointDefinition, endpointName });
```

msutkowski (changes requested):
Looks good. I didn't realize how confusing having `endpoint` be both the name or the definition was until I read this 🗡️

phryneas:
Same :sweat_smile:

msutkowski (approved):

gfortaine:
@phryneas We do confirm that it solves our problem 👍 : https://codesandbox.io/s/rtk-query-demo-forked-hd08r?file=/src/app/services/posts.ts

PR #157, opened 2021-01-24
fix: baseQuery snippet in createApi documentation

tevariou:
Refers to https://github.com/rtk-incubator/rtk-query-codegen/issues/34

### DIFF
--- a/docs/api/createApi.md
+++ b/docs/api/createApi.md
@@ -30,18 +30,22 @@ The main point where you will define a service to use in your application.
 ```ts title="Simulating axios-like interceptors with a custom base query"
 const baseQuery = fetchBaseQuery({ baseUrl: '/' });
 
-function baseQueryWithReauth(arg, api) {
-  let result = baseQuery(arg, api);
+const baseQueryWithReauth: BaseQueryFn<
+  string | FetchArgs,
+  unknown,
+  FetchBaseQueryError
+> = async (args, api, extraOptions) => {
+  let result = await baseQuery(args, api, extraOptions);
   if (result.error && result.error.status === '401') {
     // try to get a new token
-    const refreshResult = baseQuery('/refreshToken');
+    const refreshResult = await baseQuery('/refreshToken', api, extraOptions);
     if (refreshResult.data) {
       // store the new token
-      dispatch(setToken(refreshResult.data));
+      api.dispatch(setToken(refreshResult.data));
       // retry the initial query
-      result = baseQuery(arg, api);
+      result = await baseQuery(args, api, extraOptions);
     } else {
-      dispatch(loggedOut());
+      api.dispatch(loggedOut());
     }
   }
   return result;


msutkowski (approved):
LGTM. Thanks @tevariou 🎉

docs/api/createApi.md
```diff
@@ -30,22 +30,23 @@ The main point where you will define a service to use in your application.
 ```ts title="Simulating axios-like interceptors with a custom base query"
 const baseQuery = fetchBaseQuery({ baseUrl: '/' });
 
-function baseQueryWithReauth(arg, api) {
-  let result = baseQuery(arg, api);
+const baseQueryWithReauth: BaseQueryFn = async (args, api, extraOptions) => {
```
phryneas, line 33: Shouldn't at least the `Args` generic argument for `BaseQueryFn` be specified in some way?

If this just wraps `baseQuery`, we should go with `BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError, {}>`

msutkowski: Good point, sorry I missed that 🤦

tevariou: I may go with `BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError>`. Eslint is complaining on using {} as a type and It is inferred.

phryneas: Yeah, that's okay as well. 
Although I have to add that that eslint rule should be burned, since it's description is very misleading :disappointed:

phryneas (approved):

gfortaine:
That's right, it looks like that these changes were needed. On our side, we are using this snippet for auth. A few comments :

- ideally, we wouldn't like to have a 401 and use an expires_at info on the token.
- we noticed that calls made through RTK Query are concurrent thus we would need to limit them through a mutex provided by a package like this one [@shopify/semaphore](https://www.npmjs.com/package/@shopify/semaphore) (to refresh the token only once, then use the new token for subsequent calls). Here is our code :

```js
import { Semaphore } from "@shopify/semaphore";

const fetchMutex = new Semaphore(1);

const baseQueryWithReauth: BaseQueryFn<
  string | FetchArgs,
  unknown,
  FetchBaseQueryError
> = async (args, api, extraOptions) => {
  const permit = await fetchMutex.acquire();

  try {
    const { getState, dispatch } = api;
    const token = (getState() as RootState).auth.token;
    let result;

    if (token?.expires_at > Date.now())
      result = await baseQuery(args, api, extraOptions);

    if (
      token?.expires_at <= Date.now() ||
      (result.error && result.error.status === "401")
    ) {
      // try to get a new token
      const refreshResult = await baseQuery("/refreshToken", api, extraOptions);

      if (refreshResult.data) {
        // store the new token
        dispatch(setToken(refreshResult.data));
        // retry the initial query
        result = await baseQuery(args, api, extraOptions);
      } else {
        dispatch(loggedOut());
      }
    }

    return result;
  } finally {
    permit.release();
  }
};
```

PR #168, opened 2021-02-09
Fix: JSON.stringify body when it isJsonifiable and content-type is 'application/json'

### LINKED ISSUE #167: Incorrect behavior when "Content-Type: application/json" and body is not an object
It looks like that there is weird bug when we push a body that is not object with "Content-Type: application/json". And it seems that this is not the expected behavior : https://github.github.io/fetch/#request-body

A PR has been prepared : https://github.com/rtk-incubator/rtk-query/pull/168

gfortaine:
Fixes #167

### DIFF
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -110,7 +110,7 @@ export function fetchBaseQuery({
       config.headers.set('content-type', 'application/json');
     }
 
-    if (body && isPlainObject(body) && isJsonContentType(config.headers)) {
+    if (body && isJsonContentType(config.headers)) {
       config.body = JSON.stringify(body);
     }
 
--- a/src/fetchBaseQuery.ts
+++ b/src/fetchBaseQuery.ts
@@ -105,8 +105,11 @@ export function fetchBaseQuery({
 
     config.headers = await prepareHeaders(new Headers(stripUndefined(headers)), { getState });
 
-    // Only set the content-type to json if there is an object that is not FormData()
-    if (!config.headers.has('content-type') && body && typeof body === 'object' && typeof body.append !== 'function') {
+    // Only set the content-type to json if appropriate. Will not be true for FormData, ArrayBuffer, Blob, etc.
+    const isJsonifiable = (body: any) =>
+      typeof body === 'object' && (isPlainObject(body) || Array.isArray(body) || typeof body.toJSON === 'function');
+
+    if (!config.headers.has('content-type') && isJsonifiable(body)) {
       config.headers.set('content-type', 'application/json');
     }
 
--- a/test/fetchBaseQuery.test.tsx
+++ b/test/fetchBaseQuery.test.tsx
@@ -134,6 +134,24 @@ describe('fetchBaseQuery', () => {
       expect(request.body).toEqual(data);
     });
 
+    test('an array provided to body will be serialized when content-type is json', async () => {
+      const data = ['test', 'value'];
+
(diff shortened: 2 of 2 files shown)


src/fetchBaseQuery.ts
```diff
@@ -110,7 +110,7 @@ export function fetchBaseQuery({
       config.headers.set('content-type', 'application/json');
     }
 
-    if (body && isPlainObject(body) && isJsonContentType(config.headers)) {
+    if (body && isJsonContentType(config.headers)) {
```
phryneas, line 113: In all the cases mentioned on https://github.github.io/fetch/#request-body we explicitly do not want to JSON.stringify, but keep the original value, since fetch handles those natively.
Removing the `isPlainObject` check would also stringify those - maybe check if `Array.isArray` OR `isPlainObject` apply here instead?

phryneas (changes requested):

src/fetchBaseQuery.ts
gfortaine: No, because to keep the original value, it looks like that we also need to set the correct `Content-Type` header. By the way, seems that the popular [ky](https://github.com/sindresorhus/ky) is doing this way (JSON.stringify), I double-checked before submitting this PR :

https://github.com/sindresorhus/ky/blob/main/index.js#L247

```js
this._options.body = JSON.stringify(this._options.json);

Comment (phryneas) [reply src/fetchBaseQuery.ts:113]: Sorry it took so long to respond, I'm quite busy atm.

I'm not 100% in agreement here - JSON.stringifying a `FormData` will give you `{}` regardless of the contents for example, so that's pretty much never what you want.

Also, this is the code of axios: https://github.com/axios/axios/blob/master/lib/defaults.js#L31 - so the way of `ky` doing it is not really the "default way" to handle this (given that axios has 60x the downloads of ky)



So maybe we should not use the content-type of `application/json` per default, but only for plain objects/arrays and leave it untouched otherwise?

@msutkowski if you have the time - do you have an opinion on this?

Comment (msutkowski) [reply src/fetchBaseQuery.ts:113]: Sorry about the forever delay here. Looks like the `isPlainObject` was an oversight on my end - here's what I propose:

```ts
    const isJsonifiable = (body: any) => body && typeof body === 'object' && typeof body.append !== 'function';

    // Only set the content-type to json if there is an object that is not FormData()
    if (!config.headers.has('content-type') && isJsonifiable(body)) {
      config.headers.set('content-type', 'application/json');
    }

    if (isJsonifiable(body) && isJsonContentType(config.headers)) {
      config.body = JSON.stringify(body);
    }
```

msutkowski: Alright, @phryneas is right here that we don't do thorough checking, but that isn't really the goal of `fetchBaseQuery` _(imo)_. My thought process is being that we're setting the content-type automatically in some cases, it's 'safe-ish' to do what I'm proposing. If someone wanted to opt-out of that, they'd most likely be setting headers on the request anyways, which would avoid this problem. 

@gfortaine , do you have any thoughts there?

phryneas: My alternative suggestion to @msutkowski's `isJsonifiable`:

```js
 const isJsonifiable = (body: any) => typeof body === 'object' && (isPlainObject(body) || Array.isArray(body) || typeof body.toJSON === 'function');

```

msutkowski: This would be safer :)

PR #173, opened 2021-03-16
Reset promiseRef on useQuery hook

msutkowski:
Shrugsy reported this in Discord, and this solves it, but I haven't considered any other implications yet. Will review again later.

> Is there anything I can do about hot reloading causing queries to get unsubscribed and become 'uninitialized'? On a fresh CRA install, after adding a small rtk query setup, the behaviour I observe is like so:
> 1. Launch app (npm start)
> 2. Open console
> 3. Open redux devtools
> 4. [in devtools + console] Observe that data fetches normally (query fulfils in devtools, data & statuses populate in console)
> 5. [in ide] Edit and save this file (e.g. edit the logging text) to trigger a hot reload
> 6. [in devtools] Observe that the 'pokemonApi/subscriptions/unsubscribeResult' action is dispatched after the hot reload
> 7. [in console] Wait 60 seconds (or length of 'keepUnusedDataFor' option)
> 8. [in devtools] Observe that the 'pokemonApi/queries/removeQueryResult' action is dispatched
> 9. [in console] Observe that 'isUninitialized' is now back to true
> A repo that reproduces this is here: https://github.com/Shrugsy/simple-rtk-app
> Editing this file will allow you to reproduce it with the above steps: https://github.com/Shrugsy/simple-rtk-app/blob/master/src/features/pokemon/PokemonItem.jsx

### DIFF
--- a/src/react-hooks/buildHooks.ts
+++ b/src/react-hooks/buildHooks.ts
@@ -214,7 +214,11 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       ]);
 
       useEffect(() => {
-        return () => void promiseRef.current?.unsubscribe();
+        return () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          promiseRef.current?.unsubscribe();
+          promiseRef.current = undefined;
+        };
       }, []);
 
       return useMemo(
--- a/src/react-hooks/buildHooks.ts
+++ b/src/react-hooks/buildHooks.ts
@@ -284,7 +284,13 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
 
       const promiseRef = useRef<MutationActionCreatorResult<any>>();
 
-      useEffect(() => () => void promiseRef.current?.unsubscribe(), []);
+      useEffect(() => {
+        return () => {
+          // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+          promiseRef.current?.unsubscribe();
+          promiseRef.current = undefined;
+        };
+      }, []);
 
       const triggerMutation = useCallback(
         function (args) {
--- a/src/core/buildMiddleware.ts
+++ b/src/core/buildMiddleware.ts
@@ -58,6 +58,12 @@ export function buildMiddleware<Definitions extends EndpointDefinitions, Reducer
     }
 
     if (updateSubscriptionOptions.match(action)) {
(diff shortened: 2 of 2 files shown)


phryneas:
I'm not 100% sure this is the "most correct solution".

From what I gather, this is caused by the cleanup callback being run as if the component was unmounted. But after that, the other effect does not act like the component was re-mounted.
This is because we never clean the `promiseRef` in the cleanup callback and then compare it with a shallowly memoized value and see "no change".
https://github.com/rtk-incubator/rtk-query/blob/next/src/react-hooks/buildHooks.ts#L192

So I think a better fix would be to reset the refs in the cleanup callback in
https://github.com/rtk-incubator/rtk-query/blob/next/src/react-hooks/buildHooks.ts#L216-L218

src/react-hooks/buildHooks.ts
```diff
@@ -214,7 +214,10 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
       ]);
 
       useEffect(() => {
-        return () => void promiseRef.current?.unsubscribe();
+        return () => {
+          void promiseRef.current?.unsubscribe();
```
phryneas, line 218:
```suggestion
          promiseRef.current?.unsubscribe();
```

msutkowski:
> I'm not 100% sure this is the "most correct solution".
> 
> From what I gather, this is caused by the cleanup callback being run as if the component was unmounted. But after that, the other effect does not act like the component was re-mounted.
> This is because we never clean the `promiseRef` in the cleanup callback and then compare it with a shallowly memoized value and see "no change".
> https://github.com/rtk-incubator/rtk-query/blob/next/src/react-hooks/buildHooks.ts#L192
> 
> So I think a better fix would be to reset the refs in the cleanup callback in
> https://github.com/rtk-incubator/rtk-query/blob/next/src/react-hooks/buildHooks.ts#L216-L218

I agree. I think the one counterpoint though is that if someone dispatched `updateSubscriptionOptions` after scheduling a removal in a non-hook implementation, the original solution in middleware would be a 'safety net'. Then again, middleware probably shouldn't be concerned with that? :shrug:

phryneas:
@msutkowski it is a valid concern, but I think we should catch that case over in https://github.com/rtk-incubator/rtk-query/blob/ef4208904d46fbe4fbbf505e0e8a78d5670e149b/src/core/buildInitiate.ts#L125-L127 and either do nothing or even throw an exception?

phryneas (approved):

PR #177, opened 2021-03-18
useLazyQuery

msutkowski:
Introduces `useLazyQuery`, which allows users to opt-in to calling a query with arguments.

```ts
const [fetchThing, { data, isLoading, ...rest }] = api.endpoints.getPosts.useLazyQuery({refetchOnFocus, refetchOnReconnect, pollingInterval});

<button onClick={() => fetchThing({ with: 'variable', another: 'one' })}>Do a fetch</button>
```

Once a request is made, the query will stay subscribed to any changes while mounted in the scenario that there are other components mounted with the same query. If the arg that `fetchThing` is called with changes, it will perform a new request and then behave the same way. Additionally, we _always_ unsubscribe an existing query before initiating a new one when the `trigger` is called. 

[insert visual reference :D]

### DIFF
--- a/src/react-hooks/buildHooks.ts
+++ b/src/react-hooks/buildHooks.ts
@@ -29,6 +29,7 @@ export interface QueryHooks<Definition extends QueryDefinition<any, any, any, an
   useQuery: UseQuery<Definition>;
   useLazyQuery: UseLazyQuery<Definition>;
   useQuerySubscription: UseQuerySubscription<Definition>;
+  useLazyQuerySubscription: UseLazyQuerySubscription<Definition>;
   useQueryState: UseQueryState<Definition>;
 }
 
@@ -40,12 +41,6 @@ export type UseQuery<D extends QueryDefinition<any, any, any, any>> = <R = UseQu
   arg: QueryArgFrom<D>,
   options?: UseQuerySubscriptionOptions & UseQueryStateOptions<D, R>
 ) => UseQueryStateResult<D, R> & ReturnType<UseQuerySubscription<D>>;
-
-export type UseLazyQuery<D extends QueryDefinition<any, any, any, any>> = <R = UseQueryStateDefaultResult<D>>(
-  arg: QueryArgFrom<D>,
-  options?: UseQuerySubscriptionOptions & UseQueryStateOptions<D, R>
-) => [() => void, UseQueryStateResult<D, R> & ReturnType<UseQuerySubscription<D>>];
-
 interface UseQuerySubscriptionOptions extends SubscriptionOptions {
   skip?: boolean;
   refetchOnMountOrArgChange?: boolean | number;
@@ -55,6 +50,14 @@ export type UseQuerySubscription<D extends QueryDefinition<any, any, any, any>> 
   arg: QueryArgFrom<D>,
   options?: UseQuerySubscriptionOptions
 ) => Pick<QueryActionCreatorResult<D>, 'refetch'>;
+
+export type UseLazyQuery<D extends QueryDefinition<any, any, any, any>> = <R = UseQueryStateDefaultResult<D>>(
+  options?: SubscriptionOptions & Omit<UseQueryStateOptions<D, R>, 'skip'>
+) => [(arg: QueryArgFrom<D>) => void, UseQueryStateResult<D, R>];
+
+export type UseLazyQuerySubscription<D extends QueryDefinition<any, any, any, any>> = (
+  options?: SubscriptionOptions
+) => [(args: QueryArgFrom<D>) => void, React.MutableRefObject<any>];
 
 export type QueryStateSelector<R, D extends QueryDefinition<any, any, any, any>> = (
   state: QueryResultSelectorResult<D>,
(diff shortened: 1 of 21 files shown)


src/react-hooks/buildHooks.ts
```diff
@@ -262,6 +268,23 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
     return {
       useQueryState,
       useQuerySubscription,
+      useLazyQuery(arg, options) {
+        const queryStateResults = useQueryState(arg, options);
+        const [isTriggered, setTriggered] = useState((queryStateResults as any).isSuccess);
+        const { refetch, ...querySubscriptionResults } = useQuerySubscription(arg, {
+          ...options,
+          ...(isTriggered ? {} : { skip: !isTriggered }),
```
msutkowski, line 276: This is silly. We should just override this no matter what 😂

msutkowski:
Move arg into the hook call site.

msutkowski:
Note: there is a bug in this. When the input args on a call get changed, we're currently leaving stale queries around. Need to clean this up.

src/react-hooks/buildHooks.ts
```diff
@@ -229,6 +239,78 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          refetchOnReconnect,
+          refetchOnFocus,
+          pollingInterval,
+        };
+        if (lastPromise) {
+          if (!shallowEqual(options, optionsRef.current)) {
+            lastPromise.updateSubscriptionOptions(options);
+            optionsRef.current = options;
+          }
+        }
```
phryneas, line 270:
```suggestion
        if (optionsRef.current && !shallowEqual(options, optionsRef.current)) {
          lastPromise.updateSubscriptionOptions(options);
          optionsRef.current = options;
        }
```
maybe just tie it to `optionsRef` already being set so it's not two different values kinda tracking the same thing?

phryneas: We could also think about moving a "subscriptionOptions" value into the promise over in `buildInitiate` as having access to `promiseRef.current.subscriptionOptions` as the last updated subscriptionOptions might be valuable outside of this hook in other implementations as well?

src/react-hooks/buildHooks.ts
```diff
@@ -262,6 +344,24 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+        });
+        const triggerQuery = useCallback(
+          function (args: any) {
+            setProvidedArgs(args);
+            trigger(args);
+          },
+          [trigger]
+        );
+
+        return useMemo(() => [triggerQuery, queryStateResults], [triggerQuery, queryStateResults]);
```
phryneas, line 363: I kinda also want to expose the last args, so maybe

```suggestion
        const info = useMemo(() => ({ lastArgs: promise.args }, [ promise.args ]);
        return useMemo(() => [triggerQuery, queryStateResults, info], [triggerQuery, queryStateResults, info]);
```
?

src/react-hooks/buildHooks.ts
```diff
@@ -229,6 +239,78 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          argsRef.current = undefined;
+        };
+      }, []);
+
+      const trigger = useCallback(
+        function (args: any) {
+          if (!argsRef.current) {
+            argsRef.current = args;
+          } else if (argsRef.current) {
+            // args have changed, we need to unsubscribe before creating the new subscription ref.
```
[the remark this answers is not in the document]
msutkowski: I think the only point to consider here would be: Extra noise in the dev tools if someone is inspecting the behavior. Other than that, agreed. 

The only other thing I could think about is that could be an issue with things becoming out of sync due to how the selector works in `useLazyQuery`, but it's a non-issue in my 🧠 . Being that queries are shared, we'll return whatever `result` exists while the next fetch happens. So... non issue :D

src/react-hooks/buildHooks.ts
```diff
@@ -262,6 +322,25 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          function (args: any) {
+            setProvidedArgs(args);
+            trigger(args);
+          },
+          [trigger]
+        );
+
+        const info = useMemo(() => ({ lastArgs: lastPromise?.arg }), [lastPromise]);
+        return useMemo(() => [triggerQuery, queryStateResults, info], [triggerQuery, queryStateResults, info]);
+      },
```
msutkowski, line 343: I still have a concern here. It doesn't make sense for someone to call a trigger like: `getPosts(null)`, but if they did, this would be broken :thinking: . Maybe we should use some type of key like `__rtkq/uninitialized` instead `null`? Not too sure. @phryneas @markerikson any thoughts here?

src/react-hooks/buildHooks.ts
msutkowski: Updated and used as the check :+1:

src/react-hooks/buildHooks.ts
```diff
@@ -262,6 +322,25 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+        const triggerQuery = useCallback(
+          function (args: any) {
+            setProvidedArgs(args);
+            trigger(args);
+          },
+          [trigger]
+        );
+
+        const info = useMemo(() => ({ lastArgs: lastPromise?.arg }), [lastPromise]);
+        return useMemo(() => [triggerQuery, queryStateResults, info], [triggerQuery, queryStateResults, info]);
```
phryneas, line 342:
```suggestion
        const [trigger, lastPromise] = useLazyQuerySubscription(options);
        const queryStateResults = useQueryState(lastPromise.arg, {
          ...options,
          skip: !lastPromise,
        });

        const info = useMemo(() => ({ lastArgs: lastPromise?.arg }), [lastPromise]);
        return useMemo(() => [trigger, queryStateResults, info], [trigger, queryStateResults, info]);
```
Maybe doing it like this would also resolve that problem?

msutkowski: 💯

msutkowski:
Okay, remaining issues with this:

1. doing a forceUpdate in the ref as tried in this commit might be cleanest, but not sure: https://github.com/rtk-incubator/rtk-query/pull/177/commits/53ed74982e2d7a267fb5fcd6f59b340215be8bee
2. TS41Hooks need to be updated for `useLazy${thing}Query`?

msutkowski (commented):
I think this looks good. I agree with everyone who says hooks can be hard :)

PR #178, opened 2021-03-19
run type tests in CI, 4.2 compat

phryneas:


phryneas:
definitely an improvement so let's get this in

Issue #170, opened 2021-02-17
invalidation not triggering cached rejected fetches 


Kirbyasdf:
okay it seems as though the provide/invalidates is not behaving correcting since my initial fetch fails ,

currently when a user loads the app , the autheticateQuery hook will trigger, app.js, however it will return a 401 (_as the user is not logged in)_ and rtk query will set it to rejected.... then when they login, nothing happens....however if they reload the app it will fire again and redirect them to the private route..... and then if they log out, app brings them back to public routes and then if they log in again _(without refreshing) it will work as intended ... .

now the authquery is provided by the invalidation of the login and register mutations, however I believe that it only works once the user has refreshed the app is because for some reason RTK query will not refetch a failed request....

I believe that the failed 401 request is being cached and therefor even with the hook being mounted and provided by login it will not trigger, otherwise I can not see why it would not fire when it does perfectly fine after 1 successful request

 **is there some way to fix this ?**
 
If there is no fix currently can I suggest several options to the builder? : 

1. an option for refetch irregardless of cache, or at least upon certain parameters such as  response codes etc
2. an option for refetch on any state change, globally or specific states 
3. a timeout b4 refetch.. .I know there is currently an option for a refetch, however this would take more of a load off of component based logic for fetching at certain intervals 
4. built in middleware for tracking general fetches, along the lines of what is used in the initial demo videos: which I believe was custom built
5. on each query, an option to use or not use such options as prepareHeaders from the fetchBaseQuery object or any other settings and either revert to default or their own written in the builder

for reference here is the code for the api  + comps:  app / login
[repo :](https://github.com/Kirbyasdf/React-Redux-Socket-SQL-NODE)

src/services 
**api.js**

```js
import { createApi, fetchBaseQuery } from "@rtk-incubator/rtk-query";
import { RootState } from "../redux/store";

const { REACT_APP_BASE_URL } = process.env;

const baseQuery = fetchBaseQuery({
	baseUrl: REACT_APP_BASE_URL,
	prepareHeaders: (headers) => {
		const token = RootState().auth?.token;
		if (token) {
			headers.set("authorization", `Bearer ${token}`);
		}
		return headers;
	},
});

export const api = createApi({
	reducerPath: "authQuery",
	baseQuery,
	entityTypes: ["Auth"],
	endpoints: (builder) => ({
		authenticate: builder.query({
			query: () => `/auth`,
			provides: ["Auth"],
		}),
		login: builder.mutation({
[...]

PR #180, opened 2021-03-20
also call provides/invalidates in error cases

phryneas:
This adresses #170.

It is at the moment of writing a **breaking change** as it changes the signature of invalidates/provides from `(result, args) => Entities` to `(result?, error?, args) => Entities`.

It only triggers on **handled errors**, which means a simple `throw` in the `baseQuery` would pass by this, only an error returned by returning `{ error: ... }` from `baseQuery` would lead to this.

### DIFF
--- a/src/endpointDefinitions.ts
+++ b/src/endpointDefinitions.ts
@@ -1,6 +1,6 @@
 import { AnyAction, ThunkDispatch } from '@reduxjs/toolkit';
 import { RootState } from './core/apiState';
-import { BaseQueryExtraOptions, BaseQueryFn, BaseQueryResult, BaseQueryArg } from './baseQueryTypes';
+import { BaseQueryExtraOptions, BaseQueryFn, BaseQueryResult, BaseQueryError, BaseQueryArg } from './baseQueryTypes';
 import { fetchBaseQuery } from './fetchBaseQuery';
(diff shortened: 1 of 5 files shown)


src/endpointDefinitions.ts
```diff
@@ -24,16 +24,17 @@ export enum DefinitionType {
   mutation = 'mutation',
 }
 
-type GetResultDescriptionFn<EntityTypes extends string, ResultType, QueryArg> = (
-  result: ResultType,
+type GetResultDescriptionFn<EntityTypes extends string, ResultType, QueryArg, ErrorType> = (
+  result: ResultType | undefined,
+  error: ErrorType | undefined,
   arg: QueryArg
 ) => ReadonlyArray<EntityDescription<EntityTypes>>;
```
phryneas, line 31: Up for discussion: instead of going `(result, arg) => Entities` to `(result?, error?, arg) => Entities`, we could also go for

* `(arg, result?, error?) => Entities`

or

* `({result, error, arg}) => Entities`

msutkowski: i think having the original input last makes sense so it follows the whole `acc, curr, array` type of signature?

msutkowski:
Mergged in #201

PR #190, opened 2021-03-27
Add `utils.invalidateEntities`

msutkowski:
This needs some help regarding the types because they're currently a :potato:.

Allows a user to invalidate entities at any point without firing off a mutation or calling refetch on a specific endpoint. Supports `['Entity', { type: 'OtherEntity' }, { type: 'EntityWithId', id: 1 }]`

I also thought it'd be cool if we could do a `const invalidate = useInvalidateEntities();`, but I'm not sure of how to get the `EntityTypes` there without some magic.

### DIFF
--- a/src/core/module.ts
+++ b/src/core/module.ts
@@ -2,7 +2,7 @@
  * Note: this file should import all other files for type discovery and declaration merging
  */
 import { buildThunks, PatchQueryResultThunk, UpdateQueryResultThunk } from './buildThunks';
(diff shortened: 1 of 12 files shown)


src/core/module.ts
```diff
@@ -152,7 +161,7 @@ export const coreModule = (): Module<CoreModule> => ({
       baseQuery,
       reducerPath,
       context,
-      api,
+      api: api as any,
```
msutkowski, line 164: I couldn't resolve the TS error here, and I'm not too sure why being that the types should match all the way through.

phryneas (approved):

PR #192, opened 2021-03-29
Pass baseQuery into queryFn, add a test

msutkowski:


test/fetchBaseQuery.test.tsx
```diff
@@ -493,7 +493,7 @@ describe('fetchFn', () => {
 
     const baseQuery = fetchBaseQuery({
       baseUrl,
-      fetchFn: crossFetch,
+      fetchFn: crossFetch as any,
```
msutkowski, line 496: Not sure why this needed to change here...

src/endpointDefinitions.ts
```diff
@@ -27,7 +27,8 @@ export type BaseEndpointDefinition<QueryArg, BaseQuery extends BaseQueryFn, Resu
       queryFn(
         arg: QueryArg,
         api: BaseQueryApi,
-        extraOptions: BaseQueryExtraOptions<BaseQuery>
+        extraOptions: BaseQueryExtraOptions<BaseQuery>,
+        baseQuery: <Ret>(arg: any) => MaybePromise<QueryReturnValue<Ret, BaseQueryError<BaseQuery>>>
```
phryneas, line 31:
```suggestion
        baseQuery: (arg: Parameters<BaseQuery>[0]) => ReturnType<BaseQuery>
```

I think something like this might be more fitting here?

msutkowski: I might be missing something, but how would this work in this type of usage as shown in the test:

```ts
endpoints: (build) => ({
      getRandomUser: build.query<Post, void>({
        async queryFn(_arg: void, _queryApi, _extraOptions, fetchWithBQ) {
          // get a random post
          const { data } = await fetchWithBQ<Post>('posts/random');
          return fetchWithBQ<Post>(`/posts/${data?.id}`);
        },
      }),
  ```

If we used this type, I'm pretty sure it'd look like 
```ts
const { data } = await fetchWithBQ('posts/random') as QueryReturnValue<Post>`
```

phryneas: well, we *are* really just passing baseQuery without the second & third argument in here - your code from above:
```js
          (arg) => baseQuery(arg, baseQueryApi, endpointDefinition.extraOptions as any) as any
```

phryneas: But I see your point. Hmm.
Personally I'd prefer to force people to do the explicit cast as many people think that specifying a generic seems to invoke some kind of magical runtime check.

So...
```ts
const result = await fetchWithBQ('posts/random')
if (result.error) {
  throw "..."
}
const data = result.data as Post
```

msutkowski: Thanks for the clarification! Yea, that makes sense :)

phryneas (approved):
the firebase test looks good :fire:

PR #193, opened 2021-03-31
typo

dutzi:


msutkowski (approved):
@dutzi Thanks for all the typo PRs! Do you have any thoughts on the docs as a whole at this point? Any ways that we can improve? Thank you!

dutzi:
@msutkowski Sure, my pleasure!

The docs are pretty solid. I don't remember spending too much time "fighting" with RTK Query. I had one issue which took about an hour to fix:

Using GraphQL, where it's common to pass an object as the query params. I learnt, the "hard" way, that passing an object to one of the `useMeowMeowQuery`-ies is not enough, and you also have provide a params-serialization function to `createApi`. I think a simple example, or a note about that would be great. 

I'll see if I can come up with more ideas for enhancing the docs and update.

dutzi:
@msutkowski, I'm not in front of a computer right now, but IIRC the docs' approach is to provide the types through generic "parameters". 

I find the inferred approach below a bit easier to grasp, I think it's more than just an opinion, I think having the types inside the functions themselves helps the reader make the correct mapping between the types and their usage:

```ts
export const countersApi = createApi({
  baseQuery: graphqlBaseQuery(),
  entityTypes: ['Counters'],
  reducerPath: 'countersApi',
  serializeQueryArgs,
  endpoints: (builder) => ({
    listCounters: builder.query({
      query: (variables: API.ListCountersQueryVariables) => ({
        body: queries.listCounters,
        variables: variables,
      }),
      provides: listOf('Counters'),
      transformResponse: (response: API.ListCountersQuery) => response.listCounters?.items,
    }),
  }),
});
```

dutzi:
`listOf` seen above is:

```ts
function listOf<T extends string>(entity: T) {
  return (result?: ({ id: string } | null)[] | null | undefined) => [
    ...(!!result ? result!.map((item) => ({ type: entity, id: item!.id })) : []),
    { type: entity, id: 'LIST' },
  ];
}
```

Which I found pretty useful 🙂.

phryneas:
> @msutkowski, I'm not in front of a computer right now, but IIRC the docs' approach is to provide the types through generic "parameters".
> 
> I find the inferred approach below a bit easier to grasp, I think it's more than just an opinion, I think having the types inside the functions themselves helps the reader make the correct mapping between the types and their usage:
> 
> ```ts
> export const countersApi = createApi({
>   baseQuery: graphqlBaseQuery(),
>   entityTypes: ['Counters'],
>   reducerPath: 'countersApi',
>   serializeQueryArgs,
>   endpoints: (builder) => ({
>     listCounters: builder.query({
>       query: (variables: API.ListCountersQueryVariables) => ({
>         body: queries.listCounters,
>         variables: variables,
>       }),
>       provides: listOf('Counters'),
>       transformResponse: (response: API.ListCountersQuery) => response.listCounters?.items,
>     }),
>   }),
> });
> ```

Yup, that works well, too (we actually have type tests to ensure it works), but it only works if you provide `transformResponse` which many people don't really need - so we show the generic version in the docs.

PR #197, opened 2021-04-06
useQuery: skip `uninitialized` status

phryneas:
Wrapper around `defaultQueryStateSelector` to be used in `useQuery`.

 We want the initial render to already come back with  `{ isUninitialized: false, isFetching: true, isLoading: true }` to prevent that the library user has to do an additional check for `isUninitialized`/


phryneas:
Funny enough it seems we had no test that triggered on this behavioural change :tada:

msutkowski (approved):
🏆

PR #203, opened 2021-04-15
Miscellaneous documentation work

Shrugsy:
### Changelog:
- Consolidate prettier config into single `.prettierrc` file
- Miscellaneous minor updates to existing docs pages
- Move portions of docs to docblock links
- Add docblocks to various items
- Update `remark-typescript-tools`

Little is actually proposed to change in the docs due to this PR, mainly things like:
- some spelling fixes
- some consistency fixes (e.g. updating type defs in the docs to match that in the code)

Where feasible, content that *could* be an imported docblock was made into an imported docblock.

Unless I missed some, all publicly exported items *excluding* typedefs should now have a docblock (and hence show additional info when hovered in the IDE). Type defs still need covering in the future.

Note: The intent was to introduce [prettier-plugin-jsdoc](https://github.com/hosseinmd/prettier-plugin-jsdoc) to allow prettier to format examples in docblocks, but it had some small quirks with how it formatted the rest of the docblock that made me decide to forego it at this stage.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,3 +1,15 @@
 {
-  "endOfLine": "auto"
+  "endOfLine": "auto",
+  "printWidth": 120,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "overrides": [
+    {
+      "files": "*.ts",
+      "options": {
+        "parser": "babel-ts"
+      }
+    }
+  ]
 }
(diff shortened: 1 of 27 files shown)


Shrugsy:
FYI:

For docblock `@example` tags that have a 'title' in the snippet, they retain the formatting in vscode on hover, but just lose the syntax highlighting:  
image

However in codesandbox, the formatting is lost too:  
image

Up for discussion: It may be better to not link to examples where a snippet title is desired, and just duplicate the example in the docs.

PR #204, opened 2021-04-15
TS Compat tinkering

msutkowski:
Progress on fixing type tests.

- The cross fetch import was what caused the merge to blow up, and I'm not exactly sure why.


phryneas:
got it solved. how weird. :confused:

PR #191, opened 2021-03-27
WIP & Discussion: docs structure

phryneas:
What I did here was to split up the `createApi` API doc return value page into a whole own section. We need to document all these methods individually.

I've already talked to @msutkowski  and he'll take over from here (:trophy:) while I dig into the types for the other PRs :smile:  

@markerikson Any feedback on this would be highly appreciated :)

PR #206, opened 2021-04-16
selectFromResult for mutations

msutkowski:
- Adds `selectFromResult` option for `useMutation()`. 
  - This allows a user to 'optimize renders' by _only_ rerendering when the selected values change. There will always be 1 rerender when the mutation is called as it creates a new `requestId`. For the most performant implementation, you can `yourEndpoint.useMutation(undefined, { selectFromResult: () => ({}) })`
  - All possible usages can be seen in the tests with the appropriate assertions :sweat_smile:

### DIFF
--- a/src/core/apiState.ts
+++ b/src/core/apiState.ts
@@ -132,9 +132,9 @@ export type MutationSubState<D extends BaseEndpointDefinition<any, any, any>> =
   | (({
       status: QueryStatus.fulfilled;
     } & WithRequiredProp<BaseMutationSubState<D>, 'data' | 'fulfilledTimeStamp'>) & { error: undefined })
-  | ({
+  | (({
       status: QueryStatus.pending;
-    } & BaseMutationSubState<D>)
+    } & BaseMutationSubState<D>) & { data?: undefined })
   | ({
       status: QueryStatus.rejected;
     } & WithRequiredProp<BaseMutationSubState<D>, 'error'>)
--- a/src/core/buildSelectors.ts
+++ b/src/core/buildSelectors.ts
@@ -53,6 +53,10 @@ type MutationResultSelector<Definition extends MutationDefinition<any, any, any,
   requestId: string | typeof skipSelector
 ) => (state: RootState) => MutationSubState<Definition> & RequestStatusFlags;
 
+export type MutationResultSelectorResult<
+  Definition extends MutationDefinition<any, any, any, any>
+> = MutationSubState<Definition> & RequestStatusFlags;
+
 const initialSubState = {
   status: QueryStatus.uninitialized as const,
 };
--- a/src/core/buildSelectors.ts
+++ b/src/core/buildSelectors.ts
@@ -51,7 +51,7 @@ export type QueryResultSelectorResult<
 
 type MutationResultSelector<Definition extends MutationDefinition<any, any, any, any>, RootState> = (
   requestId: string | typeof skipSelector
-) => (state: RootState) => MutationSubState<Definition> & RequestStatusFlags;
+) => (state: RootState) => MutationResultSelectorResult<Definition>;
 
 export type MutationResultSelectorResult<
   Definition extends MutationDefinition<any, any, any, any>
--- a/src/react-hooks/buildHooks.ts
+++ b/src/react-hooks/buildHooks.ts
@@ -4,7 +4,6 @@ import {
   MutationSubState,
   QueryStatus,
   QuerySubState,
-  RequestStatusFlags,
   SubscriptionOptions,
   QueryKeys,
   RootState,
@@ -141,6 +140,8 @@ export type UseMutationStateOptions<D extends MutationDefinition<any, any, any, 
   selectFromResult?: MutationStateSelector<R, D>;
 };
 
+export type UseMutationStateResult<_ extends MutationDefinition<any, any, any, any>, R> = NoInfer<R>;
+
 type UseMutationStateBaseResult<D extends MutationDefinition<any, any, any, any>> = MutationSubState<D> & {
   /**
    * Mutation has not started yet.
@@ -180,7 +181,7 @@ export type UseMutation<D extends MutationDefinition<any, any, any, any>> = <R =
   ) => {
     unwrap: () => Promise<ResultTypeFrom<D>>;
   },
-  MutationSubState<D> & RequestStatusFlags
+  UseMutationStateResult<D, R>
 ];
 
 const defaultMutationStateSelector: DefaultMutationStateSelector<any> = (currentState) => {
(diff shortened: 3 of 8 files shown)


markerikson (approved):

msutkowski (commented):
This change will require updates in the docs that were merged: https://github.com/rtk-incubator/rtk-query/pull/191/files#diff-b2cf37354c870ebf35a4045446fae1006ca83193784282bb2066da6bba8a1f91R94, but was introduced for consistency with the other hook names

src/react-hooks/buildHooks.ts
```diff
@@ -423,8 +474,18 @@ export function buildHooks<Definitions extends EndpointDefinitions>({
...
+          createSelector([select(requestId || skipSelector)], (subState) =>
+            selectFromResult(subState, defaultMutationStateSelector)
+          ),
+        [select, requestId, selectFromResult]
+      );
+
+      const currentState = useSelector(
+        (state: RootState<Definitions, any, any>) => mutationSelector(state),
+        shallowEqual
+      );
```
phryneas, line 488:
```suggestion
      const currentState = useSelector(mutationSelector, shallowEqual);
```

src/react-hooks/buildHooks.ts
```diff
@@ -128,15 +127,67 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
...
 
-export type MutationHook<D extends MutationDefinition<any, any, any, any>> = () => [
+export type MutationStateSelector<R, D extends MutationDefinition<any, any, any, any>> = (
+  state: MutationResultSelectorResult<D>,
+  defaultMutationStateSelector: DefaultMutationStateSelector<D>
+) => R;
+
+export type DefaultMutationStateSelector<D extends MutationDefinition<any, any, any, any>> = (
+  state: MutationResultSelectorResult<D>
+) => UseMutationStateDefaultResult<D>;
```
phryneas, line 137:
```suggestion
) => MutationResultSelectorResult<D>;
```

src/react-hooks/buildHooks.ts
```diff
@@ -128,15 +127,67 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
...
+  | Override<Extract<UseMutationStateBaseResult<D>, { status: QueryStatus.uninitialized }>, { isUninitialized: true }>
+  | Override<
+      UseMutationStateBaseResult<D>,
+      | { isLoading: true; data: undefined }
+      | ({ isSuccess: true; error: undefined } & Required<
+          Pick<UseMutationStateBaseResult<D>, 'data' | 'fulfilledTimeStamp'>
+        >)
+      | ({ isError: true } & Required<Pick<UseMutationStateBaseResult<D>, 'error'>>)
+    >
+>;
```
phryneas, line 174:
```suggestion
```

src/react-hooks/buildHooks.ts
```diff
@@ -128,15 +127,67 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
...
+
+export type UseMutation<D extends MutationDefinition<any, any, any, any>> = <R = UseMutationStateDefaultResult<D>>(
```
phryneas, line 176:
```suggestion
export type UseMutation<D extends MutationDefinition<any, any, any, any>> = <R = MutationResultSelectorResult<D>>(
```

src/react-hooks/buildHooks.ts
```diff
@@ -128,15 +127,67 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
...
     arg: QueryArgFrom<D>
   ) => {
     unwrap: () => Promise<ResultTypeFrom<D>>;
   },
-  MutationSubState<D> & RequestStatusFlags
+  UseMutationStateResult<D, R>
 ];
 
+const defaultMutationStateSelector: DefaultMutationStateSelector<any> = (currentState) => {
+  return currentState as UseMutationStateDefaultResult<any>;
```
phryneas, line 188:
```suggestion
  return currentState;
```

test/buildHooks.test.tsx
```diff
@@ -1301,4 +1301,130 @@ describe('hooks with createApi defaults set', () => {
...
+      }),
+    });
+
+    const storeRef = setupApiStore(defaultApi, {
+      ...actionsReducer,
+    });
+
+    let getRenderCount: () => number = () => 0;
+
+    it('does not cause a rerender when using selectFromResult with an empty object', async () => {
```
phryneas, line 1333:
```suggestion
    it('causes no more than one rerender when using selectFromResult with an empty object', async () => {
```

phryneas (changes requested):
Already said it in chat - I think we don't need the fancy type overloading here as we don't really have a use for `isFetching` in mutations.

My change suggestions might be a bit sloppy, I'm not feeling so well :upside_down_face:

src/core/apiState.ts
```diff
@@ -132,9 +132,9 @@ export type MutationSubState<D extends BaseEndpointDefinition<any, any, any>> =
   | (({
       status: QueryStatus.fulfilled;
     } & WithRequiredProp<BaseMutationSubState<D>, 'data' | 'fulfilledTimeStamp'>) & { error: undefined })
-  | ({
+  | (({
       status: QueryStatus.pending;
-    } & BaseMutationSubState<D>)
+    } & BaseMutationSubState<D>) & { data?: undefined })
```
msutkowski, line 137: Forces the return type of `result.data` from `string | undefined` to `undefined` in the unionTypes test, which is what we want. `data` should never be a string there as a new request is `{ isLoading: true, data: undefined, error: undefined `

src/core/buildSelectors.ts
```diff
@@ -53,6 +53,10 @@ type MutationResultSelector<Definition extends MutationDefinition<any, any, any,
   requestId: string | typeof skipSelector
 ) => (state: RootState) => MutationSubState<Definition> & RequestStatusFlags;
```
phryneas, line 54:
```suggestion
) => (state: RootState) => MutationResultSelectorResult<Definition>;
```

src/react-hooks/buildHooks.ts
```diff
@@ -128,15 +120,36 @@ type UseQueryStateDefaultResult<D extends QueryDefinition<any, any, any, any>> =
...
   ) => {
     unwrap: () => Promise<ResultTypeFrom<D>>;
   },
-  MutationSubState<D> & RequestStatusFlags
+  UseMutationStateResult<D, R>
 ];
 
+const defaultMutationStateSelector: DefaultMutationStateSelector<any> = (currentState) => {
+  return currentState;
+};
```
phryneas, line 151:
```suggestion
const defaultMutationStateSelector: DefaultMutationStateSelector<any> = (currentState) => currentState;
```

phryneas (changes requested):
mini nitpicks, then this can go in :)

PR #209, opened 2021-04-17
Miscellaneous documentation work

Shrugsy:
### Changelog
- Update `remark-typescript-tools` to v1.0.5
- Update `provides/invalidates` examples as per the updated signature
- Update all `provides` examples that provide a specific Id to do so via the `arg` if available, regardless of result
- Add docblocks for query hook options
- Add docblocks for builder query/mutation methods & their options
- Use doclinks on the `createApi` page for parameter summaries & examples
- Fix a couple of instances of incorrect doclinks/formatting

### DIFF
--- a/docs/concepts/code-generation.md
+++ b/docs/concepts/code-generation.md
@@ -28,11 +28,16 @@ export const api = generatedApi.enhanceEndpoints({
   endpoints: {
     // basic notation: just specify properties to be overridden
     getPetById: {
-      provides: (result) => (result ? [{ type: 'Pet', id: result.id }] : []),
+      provides: (result, error, arg) => (result ? [{ type: 'Pet', id: arg.petId }] : []),
     },
     findPetsByStatus: {
       provides: (result) =>
-        result ? [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : [],
+        // is result available?
+        result ? 
+          // successful query
+          [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : 
+          // an error occured, but we still want to refetch this query when `{ type: 'Pet', id: 'LIST' }` is invalidated
+          [{ type: 'Pet', id: 'LIST' }],
     },
     // alternate notation: callback that gets passed in `endpoint` - you can freely modify the object here
     addPet: (endpoint) => {
--- a/docs/concepts/code-generation.md
+++ b/docs/concepts/code-generation.md
@@ -28,13 +28,11 @@ export const api = generatedApi.enhanceEndpoints({
   endpoints: {
     // basic notation: just specify properties to be overridden
     getPetById: {
-      provides: (response) => [{ type: 'Pet', id: response.id }],
+      provides: (result) => (result ? [{ type: 'Pet', id: result.id }] : []),
     },
     findPetsByStatus: {
-      provides: (response) => [
-        { type: 'Pet', id: 'LIST' },
-        ...response.map((pet) => ({ type: 'Pet' as const, id: pet.id })),
-      ],
+      provides: (result) =>
+        result ? [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : [],
     },
     // alternate notation: callback that gets passed in `endpoint` - you can freely modify the object here
     addPet: (endpoint) => {
@@ -44,7 +42,7 @@ export const api = generatedApi.enhanceEndpoints({
       invalidates: (response) => [{ type: 'Pet', id: response.id }],
     },
     deletePet: {
-      invalidates: (_, arg) => [{ type: 'Pet', id: arg.petId }],
+      invalidates: (result, error, arg) => [{ type: 'Pet', id: arg.petId }],
     },
   },
 });
--- a/docs/concepts/code-generation.md
+++ b/docs/concepts/code-generation.md
@@ -28,23 +28,23 @@ export const api = generatedApi.enhanceEndpoints({
   endpoints: {
     // basic notation: just specify properties to be overridden
     getPetById: {
-      provides: (result, error, arg) => (result ? [{ type: 'Pet', id: arg.petId }] : []),
+      provides: (result, error, arg) => [{ type: 'Pet', id: arg.petId }],
     },
     findPetsByStatus: {
       provides: (result) =>
         // is result available?
-        result ? 
-          // successful query
-          [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : 
-          // an error occured, but we still want to refetch this query when `{ type: 'Pet', id: 'LIST' }` is invalidated
-          [{ type: 'Pet', id: 'LIST' }],
+        result
+          ? // successful query
+            [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))]
+          : // an error occurred, but we still want to refetch this query when `{ type: 'Pet', id: 'LIST' }` is invalidated
+            [{ type: 'Pet', id: 'LIST' }],
     },
     // alternate notation: callback that gets passed in `endpoint` - you can freely modify the object here
     addPet: (endpoint) => {
-      endpoint.invalidates = (response) => [{ type: 'Pet', id: response.id }];
+      endpoint.invalidates = (result) => [{ type: 'Pet', id: result.id }];
     },
     updatePet: {
-      invalidates: (response) => [{ type: 'Pet', id: response.id }],
+      invalidates: (result, error, arg) => [{ type: 'Pet', id: arg.petId }],
     },
     deletePet: {
       invalidates: (result, error, arg) => [{ type: 'Pet', id: arg.petId }],
--- a/docs/concepts/mutations.md
+++ b/docs/concepts/mutations.md
@@ -125,11 +125,11 @@ For RTK Query, _entities_ are just a name that you can give to a specific collec
 
 #### Provides
 
-A _query_ can _provide_ entities to the cache. The `provides` argument can both be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
+A _query_ can _provide_ entities to the cache. The `provides` argument can either be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the third argument. Note that either result or error arguments may be undefined based on whether the query was successful or not.
 
 #### Invalidates
 
-A _mutation_ can _invalidate_ specific entities in the cache. The `invalidates` argument can both be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
+A _mutation_ can _invalidate_ specific entities in the cache. The `invalidates` argument can either be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the second argument. Note that either result or error arguments may be undefined based on whether the query was successful or not.
 
 ### Scenarios and Behaviors
 
@@ -144,7 +144,7 @@ export const api = createApi({
   endpoints: (build) => ({
     getPosts: build.query<PostsResponse, void>({
       query: () => 'posts',
-      provides: (result) => [...result.map(({ id }) => ({ type: 'Posts', id }))],
+      provides: (result) => (result ? result.map(({ id }) => ({ type: 'Posts', id })) : []),
     }),
     addPost: build.mutation<Post, Partial<Post>>({
       query: (body) => ({
@@ -156,7 +156,7 @@ export const api = createApi({
     }),
     getPost: build.query<Post, number>({
       query: (id) => `posts/${id}`,
-      provides: (result, id) => [{ type: 'Posts', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Posts', id }] : []),
     }),
   }),
 });
@@ -196,7 +196,8 @@ export const api = createApi({
   endpoints: (build) => ({
     getPosts: build.query<PostsResponse, void>({
       query: () => 'posts',
-      provides: (result) => [...result.map(({ id }) => ({ type: 'Posts', id })), { type: 'Posts', id: 'LIST' }],
+      provides: (result) =>
+        result ? [...result.map(({ id }) => ({ type: 'Posts', id })), { type: 'Posts', id: 'LIST' }] : [],
     }),
     addPost: build.mutation<Post, Partial<Post>>({
       query(body) {
@@ -210,7 +211,7 @@ export const api = createApi({
     }),
     getPost: build.query<Post, number>({
       query: (id) => `posts/${id}`,
-      provides: (_, id) => [{ type: 'Posts', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Posts', id }] : []),
     }),
   }),
 });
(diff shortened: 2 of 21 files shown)


Shrugsy:
Notes on stripping `'codeblock-meta'` comments within docblocks from the build:

I attempted this using [rollup-plugin-modify](https://www.npmjs.com/package/rollup-plugin-modify) like so:

```ts
const modifyPlugin = modify({
  find: /(\r|\n).*\* \/\/ codeblock-meta.*(\r|\n)/gm,
  replace: '',
});
```

While this *did* remove the lines from the corresponding `js` files, it didn't affect the typescript declaration files, so it ultimately didn't help, given that the intent was to remove that information when hovering items in the IDE.

I might have just been missing a piece of the puzzle that would allow it to remove it from the declaration files as well, but at this stage I'm ignoring it.

phryneas:
Yeah, that's just gonna be a bit extra magic in `post-build.js` - I can take a stab at that later.

docs/concepts/code-generation.md
```diff
@@ -28,13 +28,11 @@ export const api = generatedApi.enhanceEndpoints({
...
-      provides: (response) => [{ type: 'Pet', id: response.id }],
+      provides: (result) => (result ? [{ type: 'Pet', id: result.id }] : []),
     },
     findPetsByStatus: {
-      provides: (response) => [
-        { type: 'Pet', id: 'LIST' },
-        ...response.map((pet) => ({ type: 'Pet' as const, id: pet.id })),
-      ],
+      provides: (result) =>
+        result ? [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : [],
```
phryneas, line 35:
```suggestion
        // is result available?
        result ? 
          // successful query
          [{ type: 'Pet', id: 'LIST' }, ...result.map((pet) => ({ type: 'Pet' as const, id: pet.id }))] : 
          // an error occured, but we still want to refetch this query when `{ type: 'Pet', id: 'LIST' }` is invalidated
          [{ type: 'Pet', id: 'LIST' }],
```

This is actually a perfect example why we did this - even in an error case we want the list to later be refetched.

docs/concepts/code-generation.md
```diff
@@ -28,13 +28,11 @@ export const api = generatedApi.enhanceEndpoints({
   endpoints: {
     // basic notation: just specify properties to be overridden
     getPetById: {
-      provides: (response) => [{ type: 'Pet', id: response.id }],
+      provides: (result) => (result ? [{ type: 'Pet', id: result.id }] : []),
```
phryneas, line 31:
```suggestion
      provides: (result, error, arg) => (result ? [{ type: 'Pet', id: arg.petId }] : []),
```
this is more robust as it works on success and error

docs/concepts/mutations.md
```diff
@@ -125,11 +125,11 @@ For RTK Query, _entities_ are just a name that you can give to a specific collec
...
 
 #### Invalidates
 
-A _mutation_ can _invalidate_ specific entities in the cache. The `invalidates` argument can both be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument and the argument originally passed into the `query` method as the second argument.
+A _mutation_ can _invalidate_ specific entities in the cache. The `invalidates` argument can either be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the second argument. Note that either result or error arguments may be undefined based on whether the query was successful or not.
```
phryneas, line 132:
```suggestion
A _mutation_ can _invalidate_ specific entities in the cache. The `invalidates` argument can either be an array of `string` (such as `['Posts']`), `{type: string, id?: string|number}` or a callback that returns such an array. That function will be passed the result as the first argument, the response error as the second argument, and the argument originally passed into the `query` method as the second argument. Note that either result or error arguments may be undefined based on whether the mutation was successful or not.
```

docs/concepts/mutations.md
```diff
@@ -270,7 +271,8 @@ export const postApi = createApi({
       // Provides a list of `Posts` by `id`.
       // If any mutation is executed that `invalidate`s any of these entities, this query will re-run to be always up-to-date.
       // The `LIST` id is a "virtual id" we just made up to be able to invalidate this query specifically if a new `Posts` element was added.
-      provides: (result) => [...result.map(({ id }) => ({ type: 'Posts', id })), { type: 'Posts', id: 'LIST' }],
+      provides: (result) =>
+        result ? [...result.map(({ id }) => ({ type: 'Posts', id })), { type: 'Posts', id: 'LIST' }] : [],
```
phryneas, line 275:
```suggestion
      // is result available?
      result ? 
        // successful query
        [...result.map(({ id }) => ({ type: 'Posts', id })), { type: 'Posts', id: 'LIST' }] : 
        // an error occured, but we still want to refetch this query when `{ type: 'Posts', id: 'LIST' }` is invalidated
        [{ type: 'Posts', id: 'LIST' }],
```

docs/concepts/mutations.md
```diff
@@ -286,7 +288,7 @@ export const postApi = createApi({
     }),
     getPost: build.query<Post, number>({
       query: (id) => `posts/${id}`,
-      provides: (_, id) => [{ type: 'Posts', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Posts', id }] : []),
```
phryneas, line 291:
```suggestion
      provides: (result, error, id) => [{ type: 'Posts', id }],
```
this is just the most robust way to do this, so let's showcase it :)

docs/concepts/queries.md
```diff
@@ -26,7 +26,7 @@ const api = createApi({
       // `result` is the server response
       onSuccess(id, queryApi, result) {},
       onError(id, queryApi) {},
-      provides: (_, id) => [{ type: 'Post', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Post', id }] : []),
```
phryneas, line 29:
```suggestion
      provides: (result, error, id) => [{ type: 'Post', id }],
```

we could also think about adding an example like

```ts
      provides: (result, error, id) => (result ? 
        [{ type: 'Post', id }] : 
        (
          'status' in error && error.status === 401) ?
          ['UNAUTHORIZED']:
          ['UNKNOWN_ERROR']
        ),
// in combination with:
      login: build.mutation<unknown, void>({
        query: () => '/login',
        invalidates: ['UNAUTHORIZED'],
      }),
      refetchErroredQueries: build.mutation<unknown, void>({
        queryFn() {
          return { data: {} };
        },
        invalidates: ['UNKNOWN_ERROR'],
      }),

```

maybe even as a recipe and linked in the invalidation chapter?

examples/react/src/app/services/posts.ts
```diff
@@ -47,10 +47,8 @@ export const postApi = createApi({
     }),
     getPosts: build.query<PostsResponse, void>({
       query: () => ({ url: 'posts' }),
-      provides: (result) => [
-        ...result.map(({ id }) => ({ type: 'Posts', id } as const)),
-        { type: 'Posts', id: 'LIST' },
-      ],
+      provides: (result) =>
+        result ? [...result.map(({ id }) => ({ type: 'Posts', id } as const)), { type: 'Posts', id: 'LIST' }] : [],
```
phryneas, line 51:
```suggestion
        result ? [...result.map(({ id }) => ({ type: 'Posts', id } as const)), { type: 'Posts', id: 'LIST' }] : [{ type: 'Posts', id: 'LIST' }],
```

examples/react/src/app/services/posts.ts
```diff
@@ -62,7 +60,7 @@ export const postApi = createApi({
     }),
     getPost: build.query<Post, number>({
       query: (id) => `posts/${id}`,
-      provides: (_, id) => [{ type: 'Posts', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Posts', id }] : []),
```
phryneas, line 63:
```suggestion
      provides: (result, error, id) => [{ type: 'Posts', id }],
```

examples/react/src/app/services/split/post.ts
```diff
@@ -14,7 +14,7 @@ export const apiWithPost = emptySplitApi.injectEndpoints({
     }),
     getPost: build.query<Post, number>({
       query: (id) => `posts/${id}`,
-      provides: (_, id) => [{ type: 'Posts', id }],
+      provides: (result, error, id) => (result ? [{ type: 'Posts', id }] : []),
```
phryneas, line 17:
```suggestion
      provides: (result, error, id) => [{ type: 'Posts', id }],
```

src/core/apiState.ts
```diff
@@ -71,8 +71,25 @@ export function getRequestStatusFlags(status: QueryStatus): RequestStatusFlags {
 }
 
 export type SubscriptionOptions = {
+  /**
+   * How frequently to automatically re-fetch data. Defaults to `0` (off).
```
phryneas, line 75:
```suggestion
   * How frequently to automatically re-fetch data. (in milliseconds) Defaults to `0` (off).
```

src/endpointDefinitions.ts
```diff
@@ -76,9 +76,18 @@ type ResultDescription<EntityTypes extends string, ResultType, QueryArg, ErrorTy
...
+  /**
+   * A method to get the current state
+   */
   getState(): RootState<any, any, ReducerPath>;
   extra: unknown;
+  /**
+   * A unique ID generated for the mutation
+   */
   requestId: string;
   context: Context;
```
phryneas, line 92:
```suggestion
  /**
  * A variable shared between `onStart`, `onError` and `onSuccess` of one request to pass data forward between them
  */
  context: Context;
```

src/endpointDefinitions.ts
```diff
@@ -197,9 +259,58 @@ export function isMutationDefinition(
...
+   *    getPost: build.query({
+   *      query: (id) => ({ url: `post/${id}` }),
+   *      // Pick out data and prevent nested properties in a hook or selector
+   *      transformResponse: (response) => response.data,
+   *      // The 2nd parameter is the destructured `queryApi`
+   *      onStart(id, { dispatch, getState, extra, requestId, context }) {},
+   *      // `result` is the server response
+   *      onSuccess(id, queryApi, result) {},
+   *      onError(id, queryApi) {},
+   *      provides: (result, error, id) => (result ? [{ type: 'Post', id }] : []),
```
phryneas, line 280:
```suggestion
   *      provides: (result, error, id) => [{ type: 'Post', id }],
```

src/endpointDefinitions.ts
```diff
@@ -197,9 +259,58 @@ export function isMutationDefinition(
...
+   *       query: ({ id, ...patch }) => ({ url: `post/${id}`, method: 'PATCH', body: patch }),
+   *       // Pick out data and prevent nested properties in a hook or selector
+   *       transformResponse: (response) => response.data,
+   *       // onStart, onSuccess, onError are useful for optimistic updates
+   *       // The 2nd parameter is the destructured `mutationApi`
+   *       onStart({ id, ...patch }, { dispatch, getState, extra, requestId, context }) {},
+   *       // `result` is the server response
+   *       onSuccess({ id }, mutationApi, result) {},
+   *       onError({ id }, { dispatch, getState, extra, requestId, context }) {},
+   *       invalidates: ['Post'],
```
phryneas, line 308:
```suggestion
   *       invalidates: (result, error, id) => [{ type: 'Post', id }],
```

src/react-hooks/buildHooks.ts
```diff
@@ -42,8 +42,49 @@ export type UseQuery<D extends QueryDefinition<any, any, any, any>> = <R = UseQu
...
+   *
+   * - **If the query has cached data:**
+   *   * The cached data **will not be used** on the initial load, and will ignore updates from any identical query until the `skip` condition is removed
+   *   * The query will have a status of `uninitialized`
+   *   * If `skip: false` is set after skipping the initial load, the cached result will be used
+   * - **If the query does not have cached data:**
+   *   * The query will have a status of `uninitialized`
+   *   * The query will not exist in the state when viewed with the dev tools
+   *   * The query will not automatically fetch on mount
+   *   * The query will not automatically run when additional components with the same query are added that do run
```
phryneas, line 61: This reads as if it is still valid for `skip: false` and doesn't really seem right as is. Can we change that somehow?

src/react-hooks/buildHooks.ts
```diff
@@ -42,8 +42,49 @@ export type UseQuery<D extends QueryDefinition<any, any, any, any>> = <R = UseQu
...
+   *     <div>
+   *       {name} - {status}
+   *     </div>
+   *   );
+   * };
+   * ```
+   */
   skip?: boolean;
+  /**
+   * Defaults to `false`. This setting allows you to control whether RTK Query will only serve a cached result, or if it should `refetch` when set to `true` or if an adequate amount of time has passed since the last successful query result.
```
phryneas, line 81:
```suggestion
   * Defaults to `false`. This setting allows you to control whether if a cached result is already available, RTK Query will only serve a cached result, or if it should `refetch` when set to `true` or if an adequate amount of time has passed since the last successful query result.
```

src/react-hooks/buildHooks.ts
```diff
@@ -132,6 +235,21 @@ export type MutationHook<D extends MutationDefinition<any, any, any, any>> = ()
...
+     *
+     * @example
+     * ```ts
+     * // codeblock-meta title="Using .unwrap"
+     * addPost({ id: 1, name: 'Example' })
+     *   .unwrap()
+     *   .then((payload) => console.log('fulfilled', payload))
+     *   .catch((error) => console.error('rejected', error));
+     * ```
+     */
```
phryneas, line 252:
```suggestion
     * @example
     * ```ts
     * // codeblock-meta title="Using .unwrap"
     * addPost({ id: 1, name: 'Example' })
     *   .unwrap()
     *   .then((payload) => console.log('fulfilled', payload))
     *   .catch((error) => console.error('rejected', error));
     * ```
     *
     * @example
     * ```ts
     * // codeblock-meta title="Using .unwrap with async await"
     * try {
     *   const result = await addPost({ id: 1, name: 'Example' })
     *   const payload = result.unwrap();
     *   console.log('fulfilled', payload)
     * } catch (error) {
     *   console.error('rejected', error);
     * }
     * ```
     */
```

docs/concepts/queries.md
Shrugsy: I'm a big fan of this idea. I'll see if I can get something together in a separate PR for that

src/endpointDefinitions.ts
Shrugsy: Thanks, have you got a description that would be suitable for `extra` too? I left that blank for now

src/endpointDefinitions.ts
```diff
@@ -76,9 +76,18 @@ type ResultDescription<EntityTypes extends string, ResultType, QueryArg, ErrorTy
...
 export interface QueryApi<ReducerPath extends string, Context extends {}> {
+  /**
+   * The dispatch method for the store
+   */
   dispatch: ThunkDispatch<any, any, AnyAction>;
+  /**
+   * A method to get the current state
+   */
   getState(): RootState<any, any, ReducerPath>;
   extra: unknown;
```
phryneas, line 87:
```suggestion
  /**
  * `extra` as provided as `thunk.extraArgument` to the `configureStore` `getDefaultMiddleware` option.
  */
  extra: unknown;
```

src/react-hooks/buildHooks.ts
Shrugsy: Sorry I'm not understanding exactly what you mean, can you elaborate a bit please?

phryneas: Okay, had to re-read it and it *does* make sense, it's just easy to get lost in the middle.

My reading path was
>    * When skip is true:
...
> If `skip: false` is set after skipping the initial load, the cached result will be used
...
> **If the query does not have cached data:**
...
>* The query will not automatically fetch on mount

By that point I had reset my brain with the mention of `skip: false` and everything after that seemed wrong. I believe that works better when not reading it as markdown in a github diff :rofl:

src/react-hooks/buildHooks.ts
Shrugsy: The async example doesn't seem quite right, should it be like so?
```ts
const payload = await addPost({ id: 1, name: 'Example' }).unwrap()
```

phryneas: Uhh, probably :sweat_smile:
I'm not 100% awake yet :rofl:

Shrugsy: Added this in manually with the above adjustment

src/core/apiState.ts
```diff
@@ -85,13 +102,31 @@ export type MutationKeys<Definitions extends EndpointDefinitions> = {
 
 type BaseQuerySubState<D extends BaseEndpointDefinition<any, any, any>> = {
   originalArgs: QueryArgFrom<D>;
```
Shrugsy, line 104: @phryneas any recommendation for a description for `originalArgs` here?

phryneas: Hmm. Maybe something like
> argument originally passed into the hook or `initiate` action call

Shrugsy: Sounds good to me, thanks 👍

Shrugsy:
@msutkowski FYI I ended up changing a few things on the `createApi` page, sorry in advance if it conflicts with your stuff

markerikson:
Going to go ahead and merge this, then use it as the basis for a second attempt at restructuring things.

PR #217, opened 2021-04-20
Fix link to hooks in getting started

StefanBRas:
Could also link directly to docs/api/created-api/hooks.md instead.


msutkowski (approved):
Thanks @StefanBRas !

Issue #215, opened 2021-04-18
"cache entry level" lifecycle


phryneas:
We could introduce a `cacheEntryContext` next to the existing `requestContext` (renamed from just `context` and add two new lifecycle callbacks: `cacheEntryAdded` and `cacheEntryCleared`.
Those two lifecycles would allow supporting stuff like graphql subscriptions or websocket streams.

```js
    getPosts: build.query({
      query: (arg) => ({ url: `posts` }), // fetches data that will then be updated by a subscription
      cacheEntryAdded(arg, { dispatch, cacheEntryContext }) {
        cacheEntryContext.stopListening = ws.listen(() => {
          // ... 
          dispatch(
            api.util.updateQueryResult('cacheEntryAdded', arg, (draft) => {
              draft.foo = ...
            })
          )
        })
      },
      cacheEntryCleared({ id }, { dispatch, cacheEntryContext }) {
        cacheEntryContext.stopListening()
      },
    }),
 ```
 
 OR we go `useEffect` here and have `cacheEntryAdded` just return a cleanup callback. Then we could skip all the `cacheEntryContext` nonsense and omit `cacheEntryCleared`
 
 
```js
    getPosts: build.query({
      query: (arg) => ({ url: `posts` }), // fetches data that will then be updated by a subscription
      cacheEntryAdded(arg, { dispatch }) {
        const stopListening = ws.listen(() => {
          // ... 
          dispatch(
            api.util.updateQueryResult('cacheEntryAdded', arg, (draft) => {
              draft.foo = ...
            })
          )
        })
        return () => {
          stopListening()
        }
      }
    }),
 ```

markerikson:
Lenz just put up a commit with hypothetical designs for three different variations on this concept: "context"-based, "callback"-based, and "promise"-based:

https://github.com/rtk-incubator/rtk-query/pull/223/files#diff-c3ac4236290f8a1c76845f190db173acaed90fec487886e4b0e90ecbf21f13b4

Thoughts?

phryneas:
This is on the integration branch right now, closing it here.

PR #223, opened 2021-04-25
cache entry lifecycle: api experiments

phryneas:
as a follow-up to #215, some different experiments on how an API could look like & be used, asking for feedback.

final PR will be against RTK repo

### DIFF
--- a/src/core/buildInitiate.ts
+++ b/src/core/buildInitiate.ts
@@ -75,15 +75,15 @@ export type MutationActionCreatorResult<D extends MutationDefinition<any, any, a
   unsubscribe(): void;
 };
 
-export function buildInitiate<InternalQueryArgs>({
+export function buildInitiate({
   serializeQueryArgs,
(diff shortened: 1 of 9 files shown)


phryneas:
Quoting myself from chat to add to a potential open discussion here:

Personally I think #3 is quite nice because it gives you the control to delay it a bit and wait for actual values from the first request to be in the store before updating a "pending" request.
Otoh, #1 could also just add more lifecycles and #2 could just trigger later for the same result.

#3 is a bit risky if the firstValueResolved promise never resolves. We could cancel it by throwing, but what if nobody listened for it? Might be a memory leak :thinking:

Or we just .catch on the promise returned by cacheEntryAdded3 itself and dispose of any rejected firstValueResolved that way while re-throwing the rest

So, kinda still a fan of #3

phryneas:
This is on the integration branch right now, closing it here.

